PR #4, opened 2021-03-04
Setup react app

ahmedhamodi:
Doing basic create-react-app setup

Can run locally with `npm start`

Should see a page similar to:
image

We will be working off this for our admin flows.


package.json
```diff
@@ -0,0 +1,38 @@
+{
+  "name": "myfirstreactapp2",
```
ahmedhamodi, line 2: going to change this to `cdc-frontend`

aaronabraham311 (approved):
Quick question: thoughts on putting package-lock.json in gitignore? Devs would need to npm i every time but would reduce repo size.

http://gph.is/24IKcSC

ahmedhamodi:
> Quick question: thoughts on putting package-lock.json in gitignore? Devs would need to npm i every time but would reduce repo size.

Comes down to deployment choice. I know with Heroku it's needed. Not sure about AWS, which is what we'll likely go with.

For now, I'll merge this to unblock people and we can revisit later when deployment gets scoped out and worked on.

PR #12, opened 2021-03-13
Amolik circle ci

amoliksingh:
Added install-lint and install-build jobs to circle-ci to check for lint errors and successful builds.
Also added a fix for CRLF vs LF line endings


ahmedhamodi (approved):
lgtm. Thanks for doing this

PR #14, opened 2021-03-14
Add env example and update readme file.

ahmedhamodi:
Updated frontend run command to `npm run start` and added a section on secrets using the .env.example file.

Local development section looks like:
image


ahmedhamodi:
@jaydhulia I think you had told me about something else wrong with the readme file but I forgot to wrote it down - what was it if you remember?

jaydhulia:
@ahmedhamodi Yeah it was about needing to do `npm install` first and adding a standardized node version to use. The other thing for this PR is, can you add instructions for steps to add a new dependency to the project? Similar to how we have it for cdc-backend.

drbalar (approved):

jaydhulia (approved):

PR #16, opened 2021-03-18
Kevin loading screen 1

andstun:
Did some preliminary work on the first loading screen. Features a button that can generate coloured balls with random locations and a starry backdrop


ahmedhamodi:
@andstun - Can you resolve the merge conflicts? Let me know if you want to work through them together.

public/Aframe/loading-screens/scene-1/scene-1.html
```diff
@@ -0,0 +1,117 @@
+<!DOCTYPE html>
```
jaydhulia, line 1: @andstun Can you move these files to the Samples folder? Since we are still discussing the repo structure, having it in the samples folder, for now, will be helpful!

andstun: That shouldn't be a problem on my end. Did Ahmed give his confirmation? He was the one who outlined how I should structure the folders for my commit. As long as we're good there I can move all the loading screen stuff to samples.

ahmedhamodi: Yeah, Jay convinced me. :) Good to go!

public/Aframe/loading-screens/scene-1/components.js
```diff
@@ -0,0 +1,114 @@
...
+        var el = this.el;
+
+        var curBalls = document.querySelectorAll("#ball");
+
+        for (let i = 0; i < curBalls.length; i++) {
+            let container = document.querySelector("#container");
+            container.removeChild(curBalls[i]);
+        }
+
+        el.addEventListener("buttondown", function () {
```
drbalar, line 83: I was reading [this document](https://aframe.io/docs/1.2.0/introduction/writing-a-component.html#handling-property-updates) today and removing existing listeners before adding a new one is something that may be impacting performance for `update` events. I am not entirely sure though so I might be wrong, I have not gotten a chance to test it out myself yet

public/Aframe/loading-screens/scene-1/components.js
```diff
@@ -0,0 +1,114 @@
...
+    /*init: function () { //not sure if this code is necessary but keeping it in just in case
+        var data = this.data;
+        var el = this.el; 
+    },*/
+
+    update: function () {
+        var data = this.data;
+        var el = this.el;
+
+        el.addEventListener("buttondown", function () {
```
drbalar, line 101: On line 83 and here on 101, we have eventlisteners with the same name, `buttondown`. Is that intentional?

public/Aframe/loading-screens/scene-1/components.js
andstun: Took a quick glance and so far looks promising. Not sure how I didn't come across this as I remember browsing through this section. Thanks for the notice

public/Aframe/loading-screens/scene-1/components.js
andstun: My intention at the time was to have the button both generate a ball and reverse the scene's gravity. So yes, that action was intentional at the time - although may change later (important to note that the scene doesn't update for some reason, even though I'm pretty sure I'm changing the attribute correctly. So it's probably an issue with the update() function)

drbalar: Oh I see, Thanks, was just curious

public/Aframe/samples/loading-scene-1/components.js
```diff
@@ -0,0 +1,114 @@
...
+            var physics = el.getAttribute("physics");
+            console.log("physics check:", physics);
+            var gravity = physics["gravity"];
+            console.log("gravity before: ", physics);
+            gravity *= -1;
+            console.log("gravity after: ", physics);
+            el.setAttribute("physics", {
+                debug: false,
+                gravity: gravity,
+            });
```
ahmedhamodi, line 111: Since this gravity logic doesn't work you can probably just remove this

ahmedhamodi (approved):
Let's clean this up a bit more - otherwise it looks good to me!

public/Aframe/samples/loading-scene-1/components.js
```diff
@@ -0,0 +1,127 @@
...
+        "red",
+        "orange",
+        "yellow",
+        "green",
+        "blue",
+        "purple",
+        "hotpink",
+    ];
+
+    container.innerHTML += `<a-sphere collision-detect id="ball" click-drag dynamic-body position="${x} ${y} ${z}" radius="0.5" color="${
```
drbalar, line 62: While this way works, I think it would be better to use javascript to build the sphere element and append to the container instead.

So instead of `container.innerHTML = ... `, you would have something like we do in this `setupEnv() method` in [this example](https://github.com/uwblueprint/cdc-backend/blob/c0ef19b59a86e6ba2788ba04673fdda609af6bba/samples/popup/popup.html#L55).

drbalar (dismissed):
@andstun I left a few comments and besides that it looks good to go.
Feel free to reach out to me if any questions.

public/Aframe/samples/loading-scene-1/components.js
```diff
@@ -0,0 +1,127 @@
...
+            let z = Math.random() * -20 + 5;
+
+            for (let i = 0; i < 10; i++) {
+                addBall(x, y, z);
+            }
+        });
+    },
+});
+
+//this function doesn't work only because I'm pretty sure you somehow can't change physics in a-scene.
```
jaydhulia, line 91: Well, the initial gravity is 0, and you are multiplying by -1, but -1 * 0 = 0, so is that why you aren't seeing any effect of changing gravity?

Also, I recall doing this, so this is definitely doable, I think you have to do it using JS and query selector to get the <a-scene> element and using that

public/Aframe/samples/loading-scene-1/components.js
```diff
@@ -0,0 +1,127 @@
+AFRAME.registerComponent("collision-detect", {
```
jaydhulia, line 1: I think this component can be removed completely right? It only seems to be adding an event listener that does a console.log

public/Aframe/samples/loading-scene-1/scene-1.html
```diff
@@ -0,0 +1,117 @@
...
+                color="#ff1500"
+                detect-button
+                add-ball
+            ></a-cylinder>
+
+            <a-entity id="rig" position="0 0 0">
+                <a-entity
+                    id="camera"
+                    camera
+                    camera-floor
```
jaydhulia, line 103: `camera-floor` is a leftover from another example I believe, so we can remove this

jaydhulia:
Is there a reason for the package-lock.json having 20k+ changes? That seems a lot 😅 @andstun 
cc @ahmedhamodi since I am not too familiar with that file

public/Aframe/samples/loading-scene-1/components.js
andstun: It's set to 0 now, but before I was working with nonzero values without any change in result. I used my exact same approach for changing the scene's background and it worked, but for some reason, physics wasn't changing. Would you by chance have an example illustrating what you mean? I agree it doesn't make sense to me that physics would be an immutable attribute, but searching online became rather fruitless.

public/Aframe/samples/loading-scene-1/components.js
andstun: Funny you mention that. I actually tried that form, good to know that's the "better" way of doing things. If I wanted to add a custom component to an entity in this way, would the syntax be `entity.setAttribute("custom-name")`? I think that was the only reason I didn't switch to using that form.

drbalar: Yep, that is the way. Here is some aframe documentation on uses for setAttribute that might help https://aframe.io/docs/1.2.0/core/entity.html#setattribute-componentname-value-propertyvalue-clobber

public/Aframe/samples/loading-scene-1/scene-1.html
```diff
@@ -0,0 +1,117 @@
+<!DOCTYPE html>
+<html>
+    <head>
+        <meta charset="utf-8" />
+        <title>Loading Room 1</title>
+        <meta name="description" content="Loading Room 1" />
+        <script src="https://aframe.io/releases/1.1.0/aframe.min.js"></script>
+        <script src="https://cdn.jsdelivr.net/gh/donmccurdy/aframe-extras@v6.1.0/dist/aframe-extras.min.js"></script>
+        <script src="//cdn.rawgit.com/donmccurdy/aframe-physics-system/v4.0.1/dist/aframe-physics-system.min.js"></script>
```
jaydhulia, line 9: This seems to be causing an _infinite_ number of error/warning messages in the console (I had it running for 5 min and had 20k+ messages before chrome crashed)

jaydhulia, line 9: This seems to be causing an infinite number of error/warning messages in the console (I had it running for 5 min and had 20k+ messages before chrome crashed)

public/Aframe/samples/loading-scene-1/scene-1.html
```diff
@@ -0,0 +1,117 @@
+<!DOCTYPE html>
+<html>
+    <head>
+        <meta charset="utf-8" />
+        <title>Loading Room 1</title>
+        <meta name="description" content="Loading Room 1" />
+        <script src="https://aframe.io/releases/1.1.0/aframe.min.js"></script>
```
jaydhulia, line 7: The physics engine looks like is broken on aframe 1.2.0, when I tried it locally, this seems to break. Have you tried using 1.2.0?
https://github.com/n5ro/aframe-physics-system/issues/187

jaydhulia (changes requested):

public/Aframe/samples/loading-scene-1/components.js
jaydhulia: You are right, I must be mistaken here. I found this issue related to it: https://github.com/n5ro/aframe-physics-system/issues/98, it does provide a workaround, but not really sure if the workaround would work for us or not. But it seems gravity can't be changed dynamically (at least not the way we are doing right now)

public/Aframe/samples/loading-scene-1/scene-1.html
andstun: You are right, it seems like geometry is no longer defined by three.js and I just pass through the floor. I am not exactly sure how to approach this - doesn't this affect everything else that wants to use physics as well?

public/Aframe/samples/loading-scene-1/scene-1.html
andstun: The latest changes should have fixed it. I believe the reason was in Dhruvin's comment where I wasn't adding sphere's to the scene the "right" way according to A-Frame

public/Aframe/samples/loading-scene-1/components.js
```diff
@@ -0,0 +1,125 @@
...
+        el.addEventListener("buttondown", function () {
+            var curBalls = document.querySelectorAll("#ball");
+
+            for (let i = 0; i < curBalls.length; i++) {
+                curBalls[i].parentNode.removeChild(curBalls[i]);
+            }
+
+            let x = Math.random() * 20 - 10;
+            let y = 0;
+            let z = Math.random() * -20 + 5;
```
drbalar, line 80: should these be moved inside the for loop on line 82 so each ball has a possibly different spot?

public/Aframe/samples/loading-scene-1/components.js
```diff
@@ -0,0 +1,125 @@
...
+            el.setAttribute(
+                "animation__button-released",
+                "property: object3D.position.y; to: -5; dir: alternate; dur: 100; startEvents: mouseup"
+            );
+        });
+    },
+});
+
+function addBall(x, y, z) {
+    let container = document.querySelector("#container");
```
drbalar, line 39: i think this is not used anywhere so can be removed?

drbalar (commented):
Just left some suggestions/questions. And one more thing i noticed.. I did not know how to explain to I recorded a video showing it.. https://www.loom.com/share/98f53a1257334e0ba4a4116c4094faa5

Otherwise, it looks good.

public/Aframe/samples/loading-scene-1/scene-1.html
```diff
@@ -0,0 +1,114 @@
...
+                detect-button
+                add-ball
+            ></a-cylinder>
+
+            <a-entity id="rig" position="0 0 0">
+                <a-entity
+                    id="camera"
+                    camera
+                    look-controls="pointerLockEnabled: true"
+                    wasd-controls="acceleration: 200"
```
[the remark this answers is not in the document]
ahmedhamodi: Um, we'll likely switch to a nav-mesh when integrating into cdc-backend with a proper room -- I think for this sample we can keep as wasd-controls. But good point

ahmedhamodi:
Thanks for the loom video. I have no clue -- will investigate for a little bit and see if I can figure out why that is happening.

public/Aframe/samples/loading-scene-1/components.js
ahmedhamodi: Added a comment explaining why it is done outside of it - this was intentional.

ahmedhamodi (approved):

PR #17, opened 2021-03-19
Test Prettier for linting html files

amoliksingh:
Removed Aframe folder from prettierignore, then ran npm run lint-fix


ahmedhamodi (approved):
lgtm

PR #18, opened 2021-03-19
Aframe Number Lock Puzzle

amoliksingh:


public/Aframe/samples/floor.js
```diff
@@ -0,0 +1,20 @@
+AFRAME.registerComponent("floor", {
```
jaydhulia, line 1: Can you remove the components that are unrelated to the numpad from this PR @amoliksingh ?

public/Aframe/samples/number_lock.html
```diff
@@ -0,0 +1,97 @@
...
+        <script src="//cdn.rawgit.com/donmccurdy/aframe-physics-system/v4.0.1/dist/aframe-physics-system.min.js"></script>
+    </head>
+    <body>
+        <a-scene
+            cursor="rayOrigin:mouse"
+            background="color: #ECECEC"
+            physics="debug: false; gravity: -9.81"
+            raycaster="objects: .link"
+        >
+            <a-box
```
jaydhulia, line 23: I think all the floor stuff and gravity/physics stuff can be removed from this PR, as we want to compartmentalize this PR to only numpad component, that can be added to our scenes once we start working on those

public/Aframe/samples/floor.js
amoliksingh: Removed

public/Aframe/samples/number_lock.html
```diff
@@ -0,0 +1,88 @@
...
+                static-body="shape: auto"
+                position="0 -0.52 0"
+                rotation="90 0 0"
+                width="20"
+                height="10"
+                depth="1"
+                floor
+            ></a-box>
+            <a-box
```
jaydhulia, line 26: All the boxes are not needed right? Can we get rid of them?

public/Aframe/samples/keypad.js
```diff
@@ -0,0 +1,33 @@
+AFRAME.registerComponent("my_keyboard_controller", {
+    init: function () {
+        this.el.addEventListener("superkeyboardinput", function (event) {
+            if (event.detail.value === "1234") {
```
jaydhulia, line 4: Instead of hardcoding 1234 in here, can you investigate how to have this in a variable that's passed in when initializing? 
Like you know how in the .HTML file we do `super-keyboard="hand: #mouseCursor;"`, similarly, I am imagining something like `my_keyboard_controller="pass: 1234"`, and then in here using that variable. This generalization will help us get the right checks in with the actual number coming from the backend.

jaydhulia (dismissed):

public/Aframe/samples/number_lock.html
```diff
@@ -0,0 +1,88 @@
...
+                color="white"
+                floor
+            ></a-box>
+            <!-- <a-entity id="keypad-popup-container" visible="true"> -->
+            <a-entity
+                id="keyboard-display"
+                position="0 1.2 -3.9"
+                rotation="0 0 0"
+                scale="3 3 3"
+                super-keyboard="hand: #mouseCursor; imagePath:./; multipleInputs:true; filters: numbers; model: numpad; align: center; maxLength: 4; label: Enter Password; labelColor: black"
```
jaydhulia, line 76: What does `hand: #mouseCursor` do?

public/Aframe/samples/number_lock.html
amoliksingh: Removed

public/Aframe/samples/number_lock.html
ahmedhamodi: Yeah I'm also not sure what this does. But it looks like you removed it so I guess it's not needed? Can you confirmed Amolik

ahmedhamodi (approved):

jaydhulia (approved):

public/Aframe/samples/number_lock.html
amoliksingh: image

PR #19, opened 2021-03-19
Kevin loading screen 2

andstun:
Initial work for loading screen 2. The least polished of the three, if any more work is to be done probably just throw miscellaneous interactables in here lol


ahmedhamodi (commented):
This is the one that is not very finished right? I don't think we'll be landing this one, so can probably close this PR unless you think otherwise! Happy to discuss.

PR #22, opened 2021-03-20
Firebase Integration

vivianliu0:
Items completed in the PR:
- Login page is integrated with firebase
- Sign up page creates a user and is saved in firebase
- Password Reset page, which sends a reset email if the one provided is valid.
- Added routing logic so admin page is only accessible by logged in user
- Integrated with Firestore database to store user information such as displayName

passwordReset

signup

admin


src/Components/Home/home.js
```diff
@@ -1,17 +1,45 @@
...
+
     return (
-        <Container component="main" maxWidth="xs">
+        <Container component={"main"} maxWidth="xs">
             <CssBaseline />
-            <div className="Home">
+            <div className={classes.page}>
                 <Typography component="h1" variant="h5">
-                    Calgary Distress Centre
+                    Calgary Distress Centre 🤠
```
jaydhulia, line 28: Noice

ahmedhamodi:
Thank you for putting up a WIP PR so we can see your current progress. :)

src/Components/Login/firebaseCredentials.js
```diff
@@ -0,0 +1,9 @@
+export const firebaseConfig = {
```
ahmedhamodi, line 1: Nice, this is a good idea. Thanks for doing this

src/Routes.js
```diff
@@ -1,17 +1,30 @@
...
+import Signup from "./Components/Login/signup";
 import Home from "./Components/Home/home";
+import Admin from "./Components/Admin/admin";
+import PasswordReset from "./Components/Login/passwordReset";
+import { UserContext } from "./Providers/UserProviders";
 
 export default function Routes() {
-    return (
+    const user = useContext(UserContext);
+    return user ? (
```
jaydhulia, line 12: When a user is logged in, and they go to any of the pages ("/" or "/login" or "/signup"), it displays the original not signed in content first, then after a couple of seconds, it switches to Admin dashboard. Is that something that can be fixed?

jaydhulia:
LGTM other than my 1 comment. Also the circleci is failing, so you may need to fix that. Nice work on this @vivianliu0!

src/Routes.js
vivianliu0: Hmm I think this is happening because the user object isn't fully loaded yet, I'm taking a look a possible ways we could fix this. One way would be to render a loader until the object is loaded

jaydhulia: Yup a loader is something I have used before and maybe handy for other parts of the project as well. @ahmedhamodi what are your thoughts on this? Don't want to be a blocker, so adding a `loader` type functionality can be an entirely different ticket, to add that type of functionality to the project in general.

ahmedhamodi: Yeah a loader/spinner is the way to go. But can be addressed later

jaydhulia: Cool, in that case, PR looks good to me!

src/firebaseCredentials.js
```diff
@@ -0,0 +1,50 @@
...
+    if (!snapshot.exists) {
+        const { email, displayName } = user;
+        try {
+            await userRef.set({
+                displayName,
+                email,
+                ...additionalData,
+            });
+        } catch (error) {
+            console.error("Error creating user document", error);
```
jaydhulia, line 33: Should we be logging errors to the console here and on line 48?

vivianliu0: oops, will remove this, I'm surprised the linters didn't pick this up

jaydhulia: Yup seems like we are allowing it based on the eslint file (https://github.com/uwblueprint/cdc-frontend/blob/master/.eslintrc#L37). Is that expected behaviour @ahmedhamodi?

ahmedhamodi: Yes console.errors are supposed to be allowed. Only log and info I think are caught by the linter, which is what I want. We can change that if the larger team disagrees though!

ahmedhamodi: Oh, info is also allowed based on the current eslint file. I'd personally vote to remove that, but honestly no one is using console.info unless they have a reason for it, in which case it is probably fine to allow. Logs are used much more often and are the ones that might leak sensitive information to a client.

jaydhulia: Cool, we can discuss that for later. For now, let's remove these errors since they are related to auth, just in case they contain anything sensitive :) @vivianliu0

vivianliu0: Sounds good, removed

jaydhulia (dismissed):

ahmedhamodi (changes requested):
Let's not merge this just yet, I just synced with Jay offline and the reset password flow wasn't tested.

I'll test later tonight when I'm done with work and approve if everything looks fine. :)

src/Components/Login/signup.js
```diff
@@ -0,0 +1,162 @@
...
+                        required
+                        fullWidth
+                        name="password"
+                        label="Password"
+                        type="password"
+                        id="password"
+                        autoComplete="current-password"
+                        onChange={(e) => setPassword(e.target.value)}
+                        error={!empty(allErrors.login)}
+                        helperText={allErrors.login}
```
ahmedhamodi, line 139: Should be allErrors.password

src/Components/Login/signup.js
```diff
@@ -0,0 +1,162 @@
...
+
+        try {
+            const { user } = await auth.createUserWithEmailAndPassword(
+                email,
+                password
+            );
+            generateUserDocument(user, { displayName });
+            errors.login = "";
+        } catch (error) {
+            errors.login = "Email or password is incorrect.";
```
ahmedhamodi, line 59: Don't think this is needed from a signup perspective

ahmedhamodi (commented):
Ran into a small bug, added comments that should address it

ahmedhamodi:
Not sure if we have control over this code wise, but the project name isn't very pretty in the reset password email lol
image

ahmedhamodi:
> Not sure if we have control over this code wise, but the project name isn't very pretty in the reset password email lol

Fixed this through the Firebase app settings, so don't worry about this.

src/Components/Login/signup.js
vivianliu0: Thanks for catching this, after fixing it to allErrors.password it now provides a more helpful message when for example a weak password is given.

src/Components/Login/signup.js
vivianliu0: Good point, removed

ahmedhamodi (approved):
LGTM! Thanks so much for doing this 🎉

PR #23, opened 2021-03-28
WIP: Kevin input text puzzles

andstun:
Started basic functionality for a text input puzzle. The answer to the riddle is "time". This is currently not case sensitive but can easily be made to be so.

What works:
- clicking on the puzzle prompt disables movement controls and allows for pure keyboard input
- a wrong answer will generate a red X and allow for multiple user inputs
- a right answer will generate a green "congratulations". I believe I created some glitch last minute where you can still input after getting it right, which I assume is undesired behaviour (it was working correctly before though). 

What still needs to be done off the top of my head:
- add a proper "submit" button (the current submit method is using the "return" or "enter" key)
- add some sort of proper time buffer between incorrect guesses? This was stated explicitly only in Amolik's ticket but I assume applies to mine as well
- adding some buffer limit to the text box (currently there is none and it can get pretty ugly)


public/Aframe/samples/text-input-puzzles/index_text_input.html
```diff
@@ -0,0 +1,92 @@
+<!DOCTYPE html>
+<html>
+    <head>
+        <meta charset="utf-8" />
+        <title>Loading Room 1</title>
```
ahmedhamodi, line 5: nit: change title to match the sample being built

public/Aframe/samples/text-input-puzzles/index_text_input.html
```diff
@@ -0,0 +1,92 @@
...
+            physics="debug: false; gravity: 0"
+        >
+            <a-assets>
+                <img id="riddle" src="assets/riddle.jpg" />
+            </a-assets>
+
+            <!--prompt-->
+            <a-box
+                id="clue"
+                ;
```
ahmedhamodi, line 27: I don't think this should be here?

public/Aframe/samples/text-input-puzzles/components.js
```diff
@@ -0,0 +1,84 @@
...
+
+var prompt = `<a-box src="#riddle" static-body position="0 12 -20" width="10" height="10" depth="1"></a-box>`;
+
+function setPrompt(e) {}
+
+function updateInput(e) {
+    var code = parseInt(e.detail.code);
+    console.log(code);
+    document.querySelector("#input").setAttribute("color", "black");
+    switch (code) {
```
ahmedhamodi, line 12: Without really going into the logic here, what's the difference between the codes (are they just different characters) and why do some use a return and some use a break? Also, are these the only codes that we need to specifically handle in a special way?

ahmedhamodi (commented):
Initial pass. Overall structure is looking good!

public/Aframe/samples/text-input-puzzles/components.js
andstun: Yep, so for example 8 corresponds to the 'delete' key and 13 corresponds to 'return'/'enter'. The delete key needed a special case so that the input text would slice off the last character, and the return needed a special case for submitting and checking. AFAIK there are no other keys that need special treatment (other than Esc, and potentially mouse click, which is WIP). However, I haven't fully stress-tested yet.

For return vs break, that's mostly because after the switch statement I have a couple of lines to update the text to whatever character has been added and then add a trailing underscore '_' which was meant to represent a typing "cursor". Cases where I use return were for when the user got the answer right or wrong - in that case, I didn't want to update the text with the user's input, but rather a red X or congratulations and return that. If this is bad style, I can fix things to either do all return's or all break's.

ahmedhamodi: No, this is very valid. Thanks for the detailed explanation, and great idea for the _ being a current character symbol!

jaydhulia:
public/Aframe/samples/text-input-puzzles/aframe-super-keyboard.min.js - this file is not used in your PR right? I think it can be removed

ahmedhamodi:
Since we're going to use the super keyboard library for both keypads and the default keyboard, we can close this PR.

@andstun - let me know if you want to reopen this, but I don't think it's needed anymore.

PR #28, opened 2021-04-16
API helper functions

ahmedhamodi:
Not tested yet - blocked for now


jaydhulia:
@ahmedhamodi What's the blocker?

ahmedhamodi:
> @ahmedhamodi What's the blocker?

Blocker wasn't the right word - prioritizing other things. Also wanted to test with the get all endpoints, but can test with a different one. Will do that later.

aaronabraham311 (approved):

PR #29, opened 2021-04-16
Scene setup and top bars

aaronabraham311:
- Added environment bar
- Modified navbar to have a home icon based on prop passed
- Design looks ugly, but functionality is all we care about right now
- Added routes for environment

Kapture 2021-04-17 at 14 50 53

### DIFF
--- a/src/Components/Admin/EnvironmentEditor/environment.js
+++ b/src/Components/Admin/EnvironmentEditor/environment.js
@@ -18,7 +18,6 @@ export default function EnvironmentEditor() {
             <Navbar home />
             <div className={classes.page}>
                 <EnvironmentBar />
-                Hello
             </div>
         </div>
     );
--- a/src/Components/Admin/EnvironmentEditor/environment.js
+++ b/src/Components/Admin/EnvironmentEditor/environment.js
@@ -10,11 +10,7 @@ const useStyles = makeStyles((theme) => ({
     },
 }));
 
-export default function EnvironmentEditor({
-    match: {
-        params: { environmentId },
-    },
-}) {
+export default function EnvironmentEditor() {
     const classes = useStyles();
 
     return (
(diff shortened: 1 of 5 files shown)


src/Components/Admin/EnvironmentEditor/environment.js
```diff
@@ -0,0 +1,25 @@
...
+
+export default function EnvironmentEditor() {
+    const classes = useStyles();
+
+    return (
+        <div>
+            <Navbar home />
+            <div className={classes.page}>
+                <EnvironmentBar />
+                Hello
```
ahmedhamodi, line 21: Can remove this line, isn't even showing in the loom video (probably needs some padding top).

Alternatively, can do a <div> below environment bar with a paddingTop that will handle the navbar + env bar heights.

src/Components/Admin/EnvironmentEditor/environmentBar.js
```diff
@@ -0,0 +1,78 @@
...
+                        >
+                            {" "}
+                            New Scene{" "}
+                        </Button>
+                    </div>
+                    <div className={classes.buttonWrapperRight}>
+                        <Button className={classes.button}> Preview </Button>
+                        <Button className={classes.button}>
+                            {" "}
+                            Share and Publish{" "}
```
ahmedhamodi, line 71: Let's not use spaces, I would prefer padding-left and padding-right attributes on classes.button

src/Components/Admin/EnvironmentEditor/environment.js
aaronabraham311: Ah forgot about this, deleted.

src/Components/Admin/EnvironmentEditor/environmentBar.js
aaronabraham311: This was added automatically by Prettier. Deleted

ahmedhamodi: Oh interesting, not sure why Prettier does that. There might be a reason - will explore later.

ahmedhamodi (approved):
Looks good.

I fixed an issue with routing that is introduced by this PR here: https://github.com/uwblueprint/cdc-frontend/pull/32

Please review, as that PR will need to be merged shortly after this PR.

PR #30, opened 2021-04-18
Empty scene setup

aaronabraham311:
- Added buttons for empty scene as described in Notion
- Design is minimal (and might I add not great) but functionality is present
- Blocked by #29 

Screen Shot 2021-04-18 at 2 22 21 PM

### DIFF
--- a/src/Components/Admin/EnvironmentEditor/environment.js
+++ b/src/Components/Admin/EnvironmentEditor/environment.js
@@ -1,5 +1,8 @@
 import React from "react";
 import { makeStyles } from "@material-ui/core/styles";
+import Button from "@material-ui/core/Button";
+import AccountBalanceIcon from "@material-ui/icons/AccountBalance";
+import AddIcon from "@material-ui/icons/Add";
 
 import Navbar from "../navbar";
 import EnvironmentBar from "./environmentBar";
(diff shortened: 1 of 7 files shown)


src/Components/Admin/EnvironmentEditor/environment.js
```diff
@@ -19,6 +47,26 @@ export default function EnvironmentEditor() {
...
+                    <div className={classes.buttonContainer}>
+                        <Button
+                            startIcon={<AccountBalanceIcon />}
+                            className={classes.button}
+                        >
+                            New Scene from Template
+                        </Button>
+                    </div>
+                </div>
+            </div>
```
ahmedhamodi, line 69: Can we load this dynamically if a prop (arbitrarily called data) is null / empty? Not exactly sure what the structure of the data prop will be, but null should be a safe thing to check against.

ahmedhamodi: Thinking a bit more, `scenes` would be a better name and then on the call that goes to this page, we can pass in scenario.scenes, which will default to an empty array if no scenes. so the conditional rendering should be based on `scenes !== []`.

aaronabraham311: Was planning on doing this for the next ticket, unless you would like it implemented now.

ahmedhamodi: Sure next ticket is fine. As long as you are aware

ahmedhamodi (approved):

PR #31, opened 2021-04-18
Escape room page

ahmedhamodi:
Added the content of the escape room page, loaded from hardcoded data.

Loom video:
https://www.loom.com/share/1c042261621e4eb7ab974c1cd0d342aa

### DIFF
--- a/src/Components/Admin/Rooms/roomcard.js
+++ b/src/Components/Admin/Rooms/roomcard.js
@@ -1,0 +1,18 @@
+import React from "react";
+import { makeStyles } from "@material-ui/core/styles";
+
+const useStyles = makeStyles(() => ({
+    card: {
+        fontSize: 16,
+    },
+}));
+
+export default function RoomCard({ key, data }) {
+    const classes = useStyles();
+
+    return (
+        <div className={classes.card} key={key}>
+            {data.name}
+        </div>
+    );
+}
--- a/src/Components/Admin/Rooms/roomcard.js
+++ b/src/Components/Admin/Rooms/roomcard.js
@@ -1,5 +1,6 @@
 import React from "react";
 import { makeStyles } from "@material-ui/core/styles";
+import defaultImage from "./defaultImage.svg";
 
 const useStyles = makeStyles(() => ({
     card: {
@@ -12,7 +13,11 @@ export default function RoomCard({ key, data }) {
 
     return (
         <div className={classes.card} key={key}>
-            {data.name}
+            <img
+                src={data.image ? data.image : defaultImage}
+                alt="Escape Room"
+            />
+            <p>{data.name}</p>
         </div>
     );
 }
--- a/src/Components/Admin/Rooms/roomcard.js
+++ b/src/Components/Admin/Rooms/roomcard.js
@@ -1,10 +1,16 @@
 import React from "react";
 import { makeStyles } from "@material-ui/core/styles";
 import defaultImage from "./defaultImage.svg";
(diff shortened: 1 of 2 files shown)


src/Components/Admin/Rooms/roomcard.js
```diff
@@ -0,0 +1,45 @@
...
+import Grid from "@material-ui/core/Grid";
+
+const useStyles = makeStyles(() => ({
+    card: {
+        fontSize: 16,
+    },
+    cardImage: {
+        width: "90%",
+        maxWidth: 400,
+        marginTop: 16,
```
aaronabraham311, line 13: I am no front-end developer so would like some clarity on this. What's your thoughts on defining widths, margins and paddings in absolute values? Will this be a hindrance if we want to make things more responsive later on?

src/Components/Admin/Rooms/roomcard.js
```diff
@@ -0,0 +1,45 @@
...
+    },
+}));
+
+export default function RoomCard({ key, data }) {
```
aaronabraham311, line 17: Do we want to define types? I am used to doing this with TypeScript but not sure what your preference is

src/Components/Admin/Rooms/rooms.js
```diff
@@ -15,14 +17,64 @@ const useStyles = makeStyles((theme) => ({
...
+                    container
+                    item
+                    xs={12}
+                    spacing={1}
+                    alignItems="center"
+                    justify="flex-start"
+                >
+                    {rooms.map(function (room) {
+                        return <RoomCard key={room.id} data={room} />;
+                    })}
```
aaronabraham311, line 76:
```suggestion
                    {rooms.map((room) =>  (
                        <RoomCard key={room.id} data={room} />;
                    ))}
```
I am a fan of arrow functions. Just looks a lot cleaner and modern (and I believe better practice). Thoughts?

aaronabraham311 (commented):
Excellent work, especially on the responsive design. Had a few comments. Will we be implementing a white background for the card name (would be a nice starter task for @andstun )

ahmedhamodi:
> Excellent work, especially on the responsive design. Had a few comments. Will we be implementing a white background for the card name (would be a nice starter task for @andstun )

Yeah that's just CSS work so was going to postpone it. But yeah, that is a good task for Kevin.

src/Components/Admin/Rooms/rooms.js
ahmedhamodi: I heard the opposite, at my current company (not going to mention by name on github publicly), we do this. Adding a level of nesting with a return statement is not preferred.

src/Components/Admin/Rooms/roomcard.js
ahmedhamodi: Since this is in React and not Typescript, we'll avoid typing for now.

src/Components/Admin/Rooms/roomcard.js
ahmedhamodi: My general rules are the margins, padding, and maxWidth/minWidth are fine to hardcode. Widths are not great to hardcode, but in this case it makes sense since it ensures the image takes up 90% of the column (100% looks bad) and should be fine with responsiveness later on.

src/Components/Admin/Rooms/rooms.js
aaronabraham311: Interesting didn't know that. Let's go with your way then.

aaronabraham311 (approved):

PR #33, opened 2021-04-19
Scene card setup and drag and drop functionality

aaronabraham311:
- Used Atlassian's react-beautiful-dnd to create drag and drop scene cards
- Did not implement handles (will push it to later ticket, very low priority)
- Implemented conditional rendering: if there are no scenes in the state, we will render the empty scene screen that was built in #30 

Environment editor with scenes (drag and drop functionality):
Kapture 2021-04-20 at 08 59 55

Environment editor without scenes:
Kapture 2021-04-20 at 09 06 58

### DIFF
--- a/src/Components/Admin/EnvironmentEditor/environment.js
+++ b/src/Components/Admin/EnvironmentEditor/environment.js
@@ -1,11 +1,13 @@
-import React from "react";
+import React, { useState } from "react";
 import { makeStyles } from "@material-ui/core/styles";
 import Button from "@material-ui/core/Button";
 import AccountBalanceIcon from "@material-ui/icons/AccountBalance";
 import AddIcon from "@material-ui/icons/Add";
+import { DragDropContext, Droppable } from "react-beautiful-dnd";
 
 import Navbar from "../navbar";
 import EnvironmentBar from "./environmentBar";
+import SceneCard from "./sceneCard";
 
 const useStyles = makeStyles((theme) => ({
     page: {
@@ -13,6 +15,9 @@ const useStyles = makeStyles((theme) => ({
     },
     container: {
         paddingTop: theme.spacing(12),
+    },
+    dragAndDropContainer: {
+        display: "flex",
     },
     emptyButtonsContainer: {
         display: "flex",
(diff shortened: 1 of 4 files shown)


src/Components/Admin/EnvironmentEditor/environment.js
```diff
@@ -40,6 +47,49 @@ const useStyles = makeStyles((theme) => ({
...
+        },
+    ]);
+
+    const reorder = ({ scenes, startIndex, endIndex }) => {
+        const result = Array.from(scenes);
+        const [removed] = result.splice(startIndex, 1);
+        result.splice(endIndex, 0, removed);
+
+        return result;
+    };
```
aaronabraham311, line 79: Should this function be defined outside or inside the component? Not sure what the best practice is here.

ahmedhamodi: Not sure either. We can keep it here for now and refactor if it will need to be used elsewhere.

src/Components/Admin/EnvironmentEditor/environment.js
```diff
@@ -48,24 +98,52 @@ export default function EnvironmentEditor() {
...
-                            className={classes.button}
+                {scenes !== undefined && scenes.length !== 0 ? (
+                    <DragDropContext onDragEnd={onDragEnd}>
+                        <Droppable
+                            droppableId="droppable"
+                            direction="horizontal"
                         >
-                            New Scene from Template
-                        </Button>
+                            {(provided) => (
```
ahmedhamodi, line 107: I haven't seen this syntax before, is it equivalent to a conditional like `{provided ? (...) : null}`?

ahmedhamodi (commented):
Small comments. Otherwise code LGTM - will test locally.

src/Components/Admin/EnvironmentEditor/environment.js
aaronabraham311: No it's an arrow function. I found it in their sample: https://codesandbox.io/s/mmrp44okvj

ahmedhamodi: Yea I guess I was just confused how React is going to render this -- like where is this provided variable coming from? Perhaps it's a dnd feature or something

ahmedhamodi (approved):
Works locally pretty well!

PR #34, opened 2021-04-21
Create environment flow

aaronabraham311:
- Added dropdown on create button
- Added generalized modal that can and should be reused for other parts of the application
- Does not add any new environment cards
    - Why did I not do this? Currently, the state of environment cards is not in the same level defined for the creation modal. Thus, in order to implement this functionality, I would need to bring the states for the environment cards to this level. I did not think it was worth the effort for this JSON method that we will probably remove in the next few weeks.
   
Kapture 2021-04-22 at 16 23 27

### DIFF
--- a/src/Components/Admin/Rooms/roomModal.js
+++ b/src/Components/Admin/Rooms/roomModal.js
@@ -36,15 +36,15 @@ export default function RoomModal({
     };
 
     const handleModalCloseClick = () => {
+        handleModalClose();
         setRoomName("");
         setRoomDescription("");
-        handleModalClose();
     };
 
     const handleModalSubmitClick = () => {
+        handleSubmit();
         setRoomName("");
         setRoomDescription("");
-        handleSubmit();
     };
 
     return (
--- a/src/Components/Admin/Rooms/roomModal.js
+++ b/src/Components/Admin/Rooms/roomModal.js
@@ -1,5 +1,5 @@
 import React from "react";
-import FormControl from "@material-ui/core/FormControl";
+import { makeStyles } from "@material-ui/core/styles";
 import TextField from "@material-ui/core/TextField";
 import Typography from "@material-ui/core/Typography";
 import Button from "@material-ui/core/Button";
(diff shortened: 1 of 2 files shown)


src/Components/Admin/Rooms/roomModal.js
```diff
@@ -0,0 +1,77 @@
...
+
+    const handleModalCloseClick = () => {
+        setRoomName("");
+        setRoomDescription("");
+        handleModalClose();
+    };
+
+    const handleModalSubmitClick = () => {
+        setRoomName("");
+        setRoomDescription("");
```
ahmedhamodi, line 46: Why are you calling these on submit (overriding the values)?

src/Components/Admin/Rooms/roomModal.js
```diff
@@ -0,0 +1,77 @@
...
+        setRoomName(event.target.value);
+    };
+
+    const handleRoomDescriptionChange = (event) => {
+        setRoomDescription(event.target.value);
+    };
+
+    const handleModalCloseClick = () => {
+        setRoomName("");
+        setRoomDescription("");
```
ahmedhamodi, line 40: Why are we calling it here as well? Doesn't make sense in either function imo

src/Components/Admin/admin.js
```diff
@@ -133,13 +153,46 @@ export default function Admin() {
...
                         aria-label="delete"
+                        onClick={handleAddButtonClick}
                     >
                         <AddIcon />
                     </IconButton>
+                    <Menu
+                        anchorEl={anchorEl}
+                        anchorOrigin={{
+                            vertical: "top",
+                            horizontal: "center",
```
ahmedhamodi, line 169: This (and lines 172-173) can be extracted to a useStyles right?

ahmedhamodi (commented):
Small comments on code pass. Will test locally.

Definitely some functionality missing, but I understand the reasoning for pushing it until later.

src/Components/Admin/Rooms/roomModal.js
aaronabraham311: Clearing the values on submit/cancel so that previous values are not stored.

src/Components/Admin/Rooms/roomModal.js
aaronabraham311: Same comment as above

src/Components/Admin/admin.js
aaronabraham311: We could do that (not sure on the logistics) but we already had this pattern implemented as part of the navbar implementation. IMO if we are repeating this pattern throughout the component then yes, it would make sense to extract to useStyles but we are only using this behaviour once.

ahmedhamodi (approved):

PR #35, opened 2021-04-24
GET environments from backend

aaronabraham311:
- Removed all JSON data for environments
- Gets environment data from backend and renders

Screen Shot 2021-04-24 at 2 44 52 PM

### DIFF
--- a/src/Components/Admin/admin.js
+++ b/src/Components/Admin/admin.js
@@ -20,7 +20,7 @@ import EscapeRooms from "./Rooms/rooms.js";
 import Assets from "./Assets/assets.js";
 import Statistics from "./Stats/stats.js";
 import RoomModal from "./Rooms/roomModal";
-import { getAllScenes } from "../../lib/endpoints";
+import { getAllScenarios } from "../../lib/endpoints";
 
 function TabPanel(props) {
     const { children, value, index, ...other } = props;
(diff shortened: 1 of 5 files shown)


src/Components/Admin/admin.js
```diff
@@ -111,6 +112,17 @@ export default function Admin() {
...
+    const [environments, setEnvironments] = React.useState([]);
+
+    const getAllEnvironments = async () => {
+        const data = await getAllScenes();
+        setEnvironments(data);
+    };
+
+    useEffect(() => {
+        getAllEnvironments();
+    }, []);
```
ahmedhamodi, line 124: Correct me if I'm wrong, but this will get called all the time right? Should we only call it when that tab of the page is opened?

src/lib/endpoints.js
```diff
@@ -0,0 +1,6 @@
+import { httpGet } from "./dataAccess";
+
+export const getAllScenes = async () => {
+    const response = await httpGet("/api/admin/v1/scenarios");
```
jaydhulia, line 4: Shouldn't this also deal with errors? If backend returns errors? Like cookie expired, or something went wrong?

src/lib/endpoints.js
```diff
@@ -0,0 +1,6 @@
+import { httpGet } from "./dataAccess";
+
+export const getAllScenes = async () => {
```
ahmedhamodi, line 3: `getAllScenarios`

src/lib/endpoints.js
aaronabraham311: Currently we just throw an error as defined [here](https://github.com/uwblueprint/cdc-frontend/blob/1fa622173324b21e72411c82e9e58bbcdb8e8df0/src/lib/dataAccess.js#L5). I could add error handling but planning on doing it for another PR.

src/Components/Admin/admin.js
aaronabraham311: Good point. Fixed.

src/lib/endpoints.js
aaronabraham311: Oops

ahmedhamodi (approved):

PR #36, opened 2021-04-25
Integrating environment creation with POST endpoint

aaronabraham311:
- Added friendly name
- Adding environment now posts to backend and updates state

Kapture 2021-04-25 at 09 59 51

### DIFF
--- a/src/Components/Admin/admin.js
+++ b/src/Components/Admin/admin.js
@@ -20,7 +20,7 @@ import EscapeRooms from "./Rooms/rooms.js";
 import Assets from "./Assets/assets.js";
 import Statistics from "./Stats/stats.js";
 import RoomModal from "./Rooms/roomModal";
-import { getAllScenarios } from "../../lib/endpoints";
+import { getAllScenarios, postScenario } from "../../lib/endpoints";
 
 function TabPanel(props) {
     const { children, value, index, ...other } = props;
@@ -139,9 +139,14 @@ export default function Admin() {
         setAnchorEl(null);
     };
 
-    // TODO: POST room details to backend
-    const handleCreateRoomSubmit = () => {
+    const handleCreateRoomSubmit = async ({
+        name,
+        description,
+        friendly_name,
+    }) => {
         setCreateModalOpen(false);
+        const resp = await postScenario({ name, description, friendly_name });
+        setEnvironments([...environments, resp.data]);
     };
 
     return (
--- a/src/lib/dataAccess.js
+++ b/src/lib/dataAccess.js
@@ -20,9 +20,7 @@ export const httpPost = (endPoint, body) => {
 
     return new Promise((resolve, reject) => {
         axios
-            .post(url, body, {
-                headers: {},
-            })
+            .post(url, body)
             .then((response) => {
                 resolve(response);
             })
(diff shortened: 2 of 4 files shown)


src/lib/dataAccess.js
```diff
@@ -15,16 +15,13 @@ export const httpGet = (endPoint) => {
...
     const url = baseURL + endPoint;
 
     return new Promise((resolve, reject) => {
         axios
             .post(url, body, {
-                headers: {
-                    Authorization: "Bearer " + accessToken,
-                    "Content-Type": "application/json",
-                },
+                headers: {},
```
aaronabraham311, line 24: Interesting issue that I found. Axios seems to set a default header for `content-type`, which the backend does not allow (only allows `x-requested-with` from what I understand). I had to manually set empty headers but perhaps this may require some backend/frontend changes. Thoughts @jaydhulia @ahmedhamodi

ahmedhamodi: I'm okay to set empty headers, but this is def more of a Jay question so will defer to him.

src/lib/endpoints.js
```diff
@@ -1,6 +1,15 @@
-import { httpGet } from "./dataAccess";
+import { httpGet, httpPost } from "./dataAccess";
 
 export const getAllScenarios = async () => {
     const response = await httpGet("/api/admin/v1/scenarios");
     return response.data.scenarios;
 };
+
+export const postScenario = async ({ name, description, friendly_name }) => {
+    const response = await httpPost("/api/admin/v1/scenario", {
```
ahmedhamodi, line 9: I think we should move `/api/admin/v1/` to a env variable so that if we change to v2 in the future, it's an easy change.

Can you also retroactively make this change to the `getAllScenarios`?

ahmedhamodi (commented):
Will approve once we resolve the empty headers comment. Otherwise, small comments from me.

src/lib/dataAccess.js
jaydhulia: We can update the backend’s allowed headers to include this header (content-type), it’s a pretty standard header. We will also need to update this PR (and the backend allowed headers) to allow CSRF token once that’s merged in

aaronabraham311: Addressed content-type with a proposed change here: https://github.com/uwblueprint/cdc-backend/pull/84

@ahmedhamodi should I address CSRF now or only when Jay merges?

src/Components/Admin/admin.js
```diff
@@ -139,9 +139,14 @@ export default function Admin() {
...
     };
 
-    // TODO: POST room details to backend
-    const handleCreateRoomSubmit = () => {
+    const handleCreateRoomSubmit = async ({
+        name,
+        description,
+        friendly_name,
+    }) => {
         setCreateModalOpen(false);
```
jaydhulia, line 147: Is error handling and character restrictions (on friendly name for example) going to be a future PR?

aaronabraham311: Yes, not a priority right now.

src/lib/dataAccess.js
ahmedhamodi: I'm actually not sure what work is needed to address CSRF - if it's minimal I say go for it now @aaronabraham311

aaronabraham311: Let's do it once Jay merges his PR because I'm not quite sure what I would need to do as of now. Would like this merged as an example since @andstun will be working on this soon

ahmedhamodi (approved):
Changes lgtm!

PR #37, opened 2021-04-26
Edit environment

aaronabraham311:
- Created endpoint functions to edit any scenario
- Modified existing modals to handle edits

Kapture 2021-04-26 at 10 10 48

### DIFF
--- a/src/Components/Admin/Rooms/roomModal.js
+++ b/src/Components/Admin/Rooms/roomModal.js
@@ -1,4 +1,4 @@
-import React from "react";
+import React, { useEffect } from "react";
 import { makeStyles } from "@material-ui/core/styles";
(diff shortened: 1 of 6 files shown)


ahmedhamodi (commented):
Code changes look good! Going to test locally before approving since there's a few edge cases I want to test.

jaydhulia:
Shouldn't we also add the ability to edit and view the other attributes? (like whether it's published)

aaronabraham311:
> Shouldn't we also add the ability to edit and view the other attributes? (like whether it's published) Or is that going to be in a different place?

Yes that will be added in later. Currently just setting up the bare bones for editing without having to do too much design. Will probably assign the additional fields to Kevin at some point.

ahmedhamodi (approved):

PR #39, opened 2021-04-29
Connect login flow to backend endpoints

ahmedhamodi:
Loom video showing the functionality:
https://www.loom.com/share/5e3e731187354a42a8ae7ce386bc69ef

Regarding the bug I mentioned, I have a ticket that I'll work on next which will route to /admin if there exists those tokens and vice versa, if you are on an admin route without those tokens, it'll route back to /login.


src/lib/dataAccess.js
```diff
@@ -1,26 +1,38 @@
...
         });
 };
 
-export const httpPost = (endPoint, body) => {
+export const httpPost = (endPoint, body, isLogin = false) => {
     const url = baseURL + endPoint;
+    const headers = {};
+    if (!isLogin) {
+        headers["X-Xsrftoken"] = getCookie("_xsrf");
+    }
```
ahmedhamodi, line 31: This logic is needed since there is no XSRF token to set for an initial login POST request.

src/lib/dataAccess.js
```diff
@@ -1,26 +1,38 @@
 import axios from "axios";
 
 const baseURL = process.env.REACT_APP_ADMIN_BACKEND_URL;
+axios.defaults.withCredentials = true;
+
+export function getCookie(name) {
+    const r = document.cookie.match("\\b" + name + "=([^;]*)\\b");
+    return r ? r[1] : undefined;
+}
```
ahmedhamodi, line 9: Common logic from online sources -- will fetch the xsrf token from the cookie

src/Components/Login/login.js
```diff
@@ -47,31 +48,54 @@ export default function Login() {
...
 
-    function handleSubmit(event) {
+    async function handleSubmit(event) {
         event.preventDefault();
         const errors = {};
 
-        auth.signInWithEmailAndPassword(email, password).catch((error) => {
-            errors.login = "Email or password is incorrect.";
-            const errorCode = error.code;
+        auth.setPersistence(Auth.Persistence.NONE);
```
aaronabraham311, line 55: What does this line do?

src/Components/Login/login.js
```diff
@@ -47,31 +48,54 @@ export default function Login() {
...
+                    case "auth/user-disabled":
+                        errors.email = "Email has been disabled.";
+                        break;
+                    case "auth/wrong-password":
+                        errors.login = "Email or password is incorrect.";
+                        break;
+                    default:
+                        errors.login = "Email or password is incorrect.";
+                        break;
+                }
```
aaronabraham311, line 95: I'm not clear what `errors` is.  Would it make sense to use an enum to store the error messages so it's easier to change?

aaronabraham311 (commented):
I'm assuming this was tested with the backend auth branch? If so, this looks pretty good to me. Just had a few questions

src/Components/Login/login.js
ahmedhamodi: Will clean up. This was consistent with the work here before, but I agree it can be done better.

src/Components/Admin/admin.js
```diff
@@ -255,9 +256,13 @@ export default function Admin() {
                     variant="contained"
                     color="primary"
                     className={classes.submit}
-                    onClick={() => {
+                    onClick={async () => {
                         history.push("/login");
                         auth.signOut();
```
ahmedhamodi, line 261: I originally tried to get rid of the auth.signOut here, but I think it is needed since it will appropriately update the userContext which our routing is based off of.

A future todo, which I've already created, is to move off userContext and instead look at the existence of the xsrf / dcc_auth tokens as an indicator if a user is logged in or not.

src/lib/dataAccess.js
```diff
@@ -1,26 +1,38 @@
 import axios from "axios";
 
 const baseURL = process.env.REACT_APP_ADMIN_BACKEND_URL;
+axios.defaults.withCredentials = true;
```
jaydhulia, line 4: Do you need this line? Each of the below requests also have `withCredentials: true`

src/Components/Login/login.js
ahmedhamodi: I forget what the default state was (perhaps local) because I believe it caused issues with needing to sign out like mentioned in this link: https://firebase.google.com/docs/auth/web/auth-state-persistence

ahmedhamodi: Changing to SESSION because we want to keep the user logged in on refresh. Good catch

src/lib/dataAccess.js
ahmedhamodi: Doesn't look like it (just tested). Will remove

aaronabraham311 (approved):
gif

ahmedhamodi:
Will merge in sync with backend PR.

PR #40, opened 2021-04-29
Hook environment editor with GET endpoints

aaronabraham311:
- Linked each environment card with respective environment editor
- Environment editor now populates with scenes pulled from backend based on environment id

Kapture 2021-04-29 at 09 46 10

### DIFF
--- a/src/Components/Admin/EnvironmentEditor/environment.js
+++ b/src/Components/Admin/EnvironmentEditor/environment.js
@@ -1,4 +1,4 @@
-import React, { useState } from "react";
+import React, { useEffect, useState } from "react";
 import { makeStyles } from "@material-ui/core/styles";
 import Button from "@material-ui/core/Button";
 import AccountBalanceIcon from "@material-ui/icons/AccountBalance";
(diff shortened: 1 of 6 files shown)


src/Components/Admin/EnvironmentEditor/environment.js
```diff
@@ -45,30 +47,38 @@ const useStyles = makeStyles((theme) => ({
...
+            const data = await Promise.all(
+                environment.scene_ids.map(async (id) => getScene(id))
+            );
+            setScenes(data);
+        };
+
+        if (environment.scene_ids) {
+            getSceneData();
+        }
+    }, [environment]);
```
aaronabraham311, line 81: I had to define `getEnvironment` and `getSceneData` in the `useEffect` to remove some warnings, but I didn't get the same warning for the `useEffect` in `admin.js`. Anyone have any ideas why this might be the case?

ahmedhamodi: Will play around with that warning stuff a little before commenting, but also this can be combined into 1 use effect right? No need for them to be handled separately as far as I can tell

ahmedhamodi (changes requested):
There's a small bug here: https://www.loom.com/share/8e1cffa329574083a1b6f7c62250ad4b

src/Components/Admin/EnvironmentEditor/environment.js
aaronabraham311: I made them separate because I was thinking about the case where we want to start editing scene order. In this case, we would be editing the `environment` state, which should only trigger a GET request for scene data, not a re-trigger for the environment as well. I also tested combining the two together but the scene data never updated. I think it's best if we keep them separate for now

ahmedhamodi (approved):

PR #42, opened 2021-05-01
Scenes tab + backend integration for get all scenes

ahmedhamodi:
As per discussion with Jay, a scenes tab is needed since:
- Scenes are a top level component just like Scenarios
- To modify a scene, I shouldn't need to find a scenario that uses it first then modify it (esp since a scene can be in multiple scenarios)

Jay had already implemented a getAllScenes endpoint so this change didn't require backend work.

Loom video: https://www.loom.com/share/60135c81417049e8874b088712a52a13


src/Components/Admin/Scenes/scenecard.js
```diff
@@ -0,0 +1,105 @@
+import React from "react";
+import { makeStyles } from "@material-ui/core/styles";
+import { useHistory } from "react-router-dom";
+import defaultImage from "./defaultImage.svg";
```
aaronabraham311, line 4: Extremely minor nit: can we separate library imports and imports from our own files? (just have an empty space or smth). Up to you though :)

src/Components/Admin/Scenes/scenecard.js
```diff
@@ -0,0 +1,105 @@
...
+    const handleMenuClick = (event) => {
+        setAnchorEl(event.currentTarget);
+    };
+
+    const handleMenuClose = () => {
+        setAnchorEl(null);
+    };
+
+    const handleCardClick = () => {
+        history.push(`/admin/scene/${data.id}`);
```
aaronabraham311, line 36: This would go to the embedded A-frame inspector right?

src/Components/Admin/Scenes/scenecard.js
```diff
@@ -0,0 +1,105 @@
...
+    };
+
+    const open = Boolean(anchorEl);
+
+    return (
```
aaronabraham311, line 41: Hmm this looks extremely similar to the previous room card component. Is there a reason why we are not reusing it?

src/Components/Admin/Scenes/scenes.js
```diff
@@ -0,0 +1,43 @@
+import React from "react";
+import CssBaseline from "@material-ui/core/CssBaseline";
+import Container from "@material-ui/core/Container";
+import Typography from "@material-ui/core/Typography";
+import { makeStyles } from "@material-ui/core/styles";
+import SceneCard from "./scenecard.js";
```
aaronabraham311, line 6: Same nit as above

aaronabraham311 (commented):
Clarified this in a discussion with Ahmed. This was not in the designs but we can add it given that there is a use case.

src/Components/Admin/Scenes/scenecard.js
ahmedhamodi: You'll need to work with the designers, but we might want some sort of parent screen with information like name, last modified, etc with a button to launch aframe inspector.

Otherwise, we can have all this info in a popup or sorts and just link this to aframe inspector right away. I'm okay with either, comes down to a design decision imo

src/Components/Admin/Scenes/scenecard.js
ahmedhamodi: We probably can, let me do that

src/Components/Admin/Scenes/scenecard.js
ahmedhamodi: Going to leave this as is, we can change it in a future PR.

aaronabraham311 (approved):
LGTM! Question I have because I am no pro in frontend development: when do we use prop types and when do we not? Do we want to include prop types?

ahmedhamodi:
> LGTM! Question I have because I am no pro in frontend development: when do we use prop types and when do we not? Do we want to include prop types?

This is a feature of typescript which we can definitely use if we want. At the moment, I'm okay with just React. Will keep this in mind for the future!

PR #43, opened 2021-05-03
Create scene from dropdown

aaronabraham311:
- Added scene dropdown
- Added scene modal
- Integrated scene creation endpoint

Kapture 2021-05-03 at 09 50 53

### DIFF
--- a/src/Components/Admin/EnvironmentEditor/environment.js
+++ b/src/Components/Admin/EnvironmentEditor/environment.js
@@ -115,11 +115,11 @@ export default function EnvironmentEditor({
         setCreateModalOpen(false);
 
         const newScene = await createScene(name, background_id);
-        const newSceneData = [...scenes, newScene.id];
+        const newSceneData = [...scenes, newScene];
         setScenes(newSceneData);
 
         const newEnvData = environment;
-        newEnvData.scene_ids = [...environment.scene_ids, newScene.id.id];
+        newEnvData.scene_ids = [...environment.scene_ids, newScene.id];
         const newEnv = await editScenario(newEnvData);
         setEnvironment(newEnv);
     };
--- a/src/Components/Admin/EnvironmentEditor/environment.js
+++ b/src/Components/Admin/EnvironmentEditor/environment.js
@@ -8,8 +8,9 @@ import { DragDropContext, Droppable } from "react-beautiful-dnd";
 import Navbar from "../navbar";
 import EnvironmentBar from "./environmentBar";
 import SceneCard from "./sceneCard";
-import { getScenario } from "../../../lib/scenarioEndpoints";
-import { getScene } from "../../../lib/sceneEndpoints";
+import SceneModal from "./sceneModal";
+import { getScenario, editScenario } from "../../../lib/scenarioEndpoints";
+import { getScene, createScene } from "../../../lib/sceneEndpoints";
 
 const useStyles = makeStyles((theme) => ({
     page: {
@@ -55,6 +56,7 @@ export default function EnvironmentEditor({
     const classes = useStyles();
     const [environment, setEnvironment] = useState({});
     const [scenes, setScenes] = useState([]);
+    const [createModalOpen, setCreateModalOpen] = useState(false);
 
     useEffect(() => {
         const getEnvironment = async () => {
@@ -101,11 +103,32 @@ export default function EnvironmentEditor({
         setScenes(items);
     };
 
+    const onCreateButtonClick = () => {
+        setCreateModalOpen(true);
+    };
+
+    const onCreateModalClose = () => {
+        setCreateModalOpen(false);
+    };
+
+    const onCreateModalSubmit = async (name, background_id) => {
+        setCreateModalOpen(false);
+
+        const newScene = await createScene(name, background_id);
+        const newSceneData = [...scenes, newScene.id];
+        setScenes(newSceneData);
+
+        const newEnvData = environment;
+        newEnvData.scene_ids = [...environment.scene_ids, newScene.id.id];
+        const newEnv = await editScenario(newEnvData);
+        setEnvironment(newEnv);
+    };
+
     return (
         <div>
             <Navbar home />
             <div className={classes.page}>
-                <EnvironmentBar />
+                <EnvironmentBar onCreateButtonClick={onCreateButtonClick} />
             </div>
             <div className={classes.container}>
                 {scenes !== undefined && scenes.length !== 0 ? (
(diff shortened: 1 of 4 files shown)


src/Components/Admin/EnvironmentEditor/environment.js
```diff
@@ -101,11 +103,32 @@ export default function EnvironmentEditor({
...
+        setCreateModalOpen(false);
+
+        const newScene = await createScene(name, background_id);
+        const newSceneData = [...scenes, newScene.id];
+        setScenes(newSceneData);
+
+        const newEnvData = environment;
+        newEnvData.scene_ids = [...environment.scene_ids, newScene.id.id];
+        const newEnv = await editScenario(newEnvData);
+        setEnvironment(newEnv);
```
aaronabraham311, line 124: This is more complicated than it needs to be. I propose that we change our backend scene POST endpoint to include the environment ID so that the environment is automatically updated rather than having to send two requests.

Furthermore, the response that I get from scene POST endpoint is a little weird, hence why I used `newScene.id`.  Not sure why the data is nested under `id`. 

Screen Shot 2021-05-03 at 9 53 12 AM

Thoughts on the above @ahmedhamodi @jaydhulia

jaydhulia: I don't think we should change the backend scene POST endpoint. The scene is meant to be separate from a scenario, so including a scenario ID with a scene creation doesn't make sense to me. Ideally, admins create scene on the scenes tab that Ahmed has created. And then on the scenario, when they want to add a scene, they simply select from already existing scenes.
That was the workflow that was kept in mind when creating the backend database schema. That makes more sense to me, because admins may want to reuse scenes across scenarios.
The scenario is simply to have an ordered collection of scenes and control whether they are published.

The other `id` issue has been addressed via Slack.

aaronabraham311: > Furthermore, the response that I get from scene POST endpoint is a little weird, hence why I used newScene.id. Not sure why the data is nested under id.

Resolved

src/Components/Admin/EnvironmentEditor/sceneModal.js
```diff
@@ -0,0 +1,88 @@
...
+    };
+
+    const handleModalCloseClick = () => {
+        handleModalClose();
+        setSceneName("");
+        setBackgroundId(null);
+    };
+
+    const handleModalSubmitClick = () => {
+        handleSubmit(sceneName, parseInt(backgroundId));
```
ahmedhamodi, line 54: Will eventually be integrating the background assets to a dropdown right? ID isn't the most convenient

ahmedhamodi (commented):
Code changes look good to me, will test locally.

src/Components/Admin/EnvironmentEditor/environment.js
ahmedhamodi: I think it's fine to expose creation of a scene in both locations personally.

It's just when it's done here, the linking to an environment needs to be explicitly done.

If done through the scenes tab, that is not needed as it is independent and not linked to any environment.

Because of this second flow that doesn't need the linkage, it should be done separate from the POST endpoint.

src/Components/Admin/EnvironmentEditor/sceneModal.js
aaronabraham311: Depends on hifi, but currently just asking for background ID submission. Will probably change in the near future.

ahmedhamodi:
Noticing this behaviour when trying names with numbers and with using a background ID that doesn't exist:
https://www.loom.com/share/070be116e96f4ddab8f09c0c091a0ffa

Thoughts?

jaydhulia:
@ahmedhamodi for that error, is it the backend that’s not returning the error properly or is it the frontend doing something weird? Aka can you share what the network request looks like?

ahmedhamodi:
> @ahmedhamodi for that error, is it the backend that’s not returning the error properly or is it the frontend doing something weird? Aka can you share what the network request looks like?

I think the issue is that the validation rule doesn't allow numbers?

```
{"status": 500, "title": "Internal Server Error", "message": "'test 123' does not match '^[a-zA-Z _-]{1,50}$'\n\nFailed validating 'pattern' in schema['properties']['name']:\n    {'pattern': '^[a-zA-Z _-]{1,50}$', 'type': 'string'}\n\nOn instance['name']:\n    'test 123'"}
```

The payload looks fine to me:
image

aaronabraham311:
> I think the issue is that the validation rule doesn't allow numbers?
This is true. Again, I have not put any sort of error handling in here as with the rest of my PRs.

jaydhulia:
I meant the second error that you showed in your loom video @ahmedhamodi , like the one where you have a React error. Is the backend not returning proper JSON or is the frontend not error handling?
The case where you created a scene named “test” with a background id that doesn’t exist.

ahmedhamodi:
> I meant the second error that you showed in your loom video @ahmedhamodi , like the one where you have a React error. Is the backend not returning proper JSON or is the frontend not error handling?
> The case where you created a scene named “test” with a background id that doesn’t exist.

Hmm, I'm not able to reproduce now. New error message:
```
{"status": 500, "title": "Internal Server Error", "message": "(psycopg2.errors.ForeignKeyViolation) insert or update on table \"scene\" violates foreign key constraint \"scene_background_id_fkey\"\nDETAIL:  Key (background_id)=(453) is not present in table \"asset\".\n\n[SQL: INSERT INTO scene (name, description, object_ids, position, scale, rotation, background_id, camera_properties, hints) VALUES (%(name)s, %(description)s, %(object_ids)s, %(position)s, %(scale)s, %(rotation)s, %(background_id)s, %(camera_properties)s, %(hints)s) RETURNING scene.id]\n[parameters: {'name': 'test', 'description': '', 'object_ids': [], 'position': [], 'scale': [], 'rotation': [], 'background_id': 453, 'camera_properties': '{}', 'hints': []}]\n(Background on this error at: http://sqlalche.me/e/13/gkpj)"}
```

Video showing me retrying:
https://www.loom.com/share/c55fa7a400084ae18760f391ce06ee26

jaydhulia:
Yeah that’s what I would expect from the backend, it will be a postgres validation error. It’s probably how it’s handled on the frontend that caused the initial error. From the backend perspective, it’s as expected.

ahmedhamodi (approved):
So this is good to land then I assume.

@jaydhulia can we allow numbers in the names of scenarios, scenes, and object/assets?

PR #44, opened 2021-05-03
Fix DOMNesting Bug

vivianliu0:
Previously, a weird validateDOMNesting warning was appearing in console when a user navigates to the main route. This error was occurring while using the Material UI `<Typography>` element, which picks `<p>` as the default root component if none is specified. 

Image of error:
image


ahmedhamodi:
Let's clean up some of the prettier issues and update the branch. But putting those aside, the code changes look good!

ahmedhamodi (approved):
lgtm - no console errors when running locally! Thanks for doing this 🎉

PR #47, opened 2021-05-07
Added styling for text and menu

andstun:
Pair programmed with Aaron to learn some React basics and change the styling on environment name cards:
Screen Shot 2021-05-06 at 9 30 44 PM

### DIFF
--- a/src/Components/Admin/common/itemCard.js
+++ b/src/Components/Admin/common/itemCard.js
@@ -17,6 +17,20 @@ const useStyles = makeStyles(() => ({
         maxWidth: 400,
         marginTop: 16,
     },
+    metadata: {
+        display: "flex",
+        justifyContent: "space-between",
+        marginLeft: "5.7%",
+        marginRight: "5.7%",
+        marginTop: -4,
+        backgroundColor: "white",
+        borderBottomLeftRadius: "5px",
+        borderBottomRightRadius: "5px",
+    },
+    dataName: {
+        marginLeft: 10,
+        fontWeight: "bold",
+    },
 }));
 
 export default function ItemCard({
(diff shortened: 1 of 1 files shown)


src/Components/Admin/common/itemCard.js
```diff
@@ -17,6 +17,20 @@ const useStyles = makeStyles(() => ({
         maxWidth: 400,
         marginTop: 16,
     },
+    metadata: {
+        display: "flex",
+        justifyContent: "space-between",
+        marginLeft: "5.7%",
+        marginRight: "5.7%",
```
andstun, line 24: @ahmedhamodi is this best practice, doing decimals with percentages?

aaronabraham311 (approved):
Pair programmed with him and things are good. Tested 2 and 3 card grids as well

src/Components/Admin/common/itemCard.js
ahmedhamodi: I think I prefer `vw` and `vh` (see https://www.w3schools.com/cssref/css_units.asp) but this should be fine since CSS is not final.

src/Components/Admin/common/itemCard.js
```diff
@@ -17,6 +17,20 @@ const useStyles = makeStyles(() => ({
...
+        marginTop: -4,
+        backgroundColor: "white",
+        borderBottomLeftRadius: "5px",
+        borderBottomRightRadius: "5px",
```
ahmedhamodi, line 28: Nit: the designs in figma appear to have a larger border radius here, similar to what is on the top left and right of the image.

ahmedhamodi (approved):

PR #48, opened 2021-05-07
Edit scenario on reorder

aaronabraham311:
- On reorder, the frontend will fire a PUT request to the backend to edit the scenario scene_ids
- Removed a warning coming from beautiful dnd

Loom: https://www.loom.com/share/70c761693e0548f186c2295ebabfe272

### DIFF
--- a/src/Components/Admin/EnvironmentEditor/environment.js
+++ b/src/Components/Admin/EnvironmentEditor/environment.js
@@ -86,11 +86,10 @@ export default function EnvironmentEditor({
         const result = Array.from(scenes);
         const [removed] = result.splice(startIndex, 1);
         result.splice(endIndex, 0, removed);
-
         return result;
     };
 
-    const onDragEnd = (result) => {
+    const onDragEnd = async (result) => {
         if (!result.destination) {
             return;
         }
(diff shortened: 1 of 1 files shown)


src/Components/Admin/EnvironmentEditor/environment.js
```diff
@@ -101,6 +100,18 @@ export default function EnvironmentEditor({
...
+        const newSceneIds = items.map((scene) => scene.id);
+        const response = await editScenario({
+            id: environment.id,
+            name: environment.name,
+            friendly_name: environment.friendly_name,
+            description: environment.description,
+            scene_ids: newSceneIds,
+            is_published: environment.is_published,
+            is_previewable: environment.is_previewable,
+        });
```
aaronabraham311, line 113: Not a fan of this, would have liked to do object destructuring like so:
```javascript
const { oldSceneIds, ...rest } = environment
const response = await editScenario({
    scene_ids: newSceneIds,
    ...rest,
})
```
Much cleaner but then I get warnings of unused variables :(

ahmedhamodi: I'm assuming the error was on `oldSceneIds` not being used?

If it's an eslint error, which I suspect it is, you can add the following line above it:
`// eslint-disable-next-line no-unused-vars`

And there should be no more warnings for that.

aaronabraham311: Ah it's ok, not too big of a deal. Let's just merge it in. If we feel like we need a cleaner way, then we can change it then. I'm fine with the way I implemented it for now.

ahmedhamodi (approved):

PR #51, opened 2021-05-10
Error Component

vivianliu0:
Error component that is displayed when we receive an error from the backend.

Loom Video for Error Component:
https://www.loom.com/share/a05d6126cc1741b28ee562e36dd37bdb

### DIFF
--- a/src/Components/Admin/EnvironmentEditor/environment.js
+++ b/src/Components/Admin/EnvironmentEditor/environment.js
@@ -62,12 +62,8 @@ export default function EnvironmentEditor({
 
     useEffect(() => {
         const getEnvironment = async () => {
-            try {
-                const data = await getScenario(environmentId);
-                setEnvironment(data);
-            } catch (error) {
-                handleError(error);
-            }
+            const data = await getScenario(environmentId, handleError);
+            setEnvironment(data);
         };
 
         if (environmentId) {
@@ -77,14 +73,12 @@ export default function EnvironmentEditor({
 
     useEffect(() => {
         const getSceneData = async () => {
-            try {
-                const data = await Promise.all(
-                    environment.scene_ids.map(async (id) => getScene(id))
-                );
-                setScenes(data);
-            } catch (error) {
-                handleError(error);
-            }
+            const data = await Promise.all(
+                environment.scene_ids.map(async (id) =>
+                    getScene(id, handleError)
+                )
+            );
+            setScenes(data);
         };
 
         if (environment.scene_ids) {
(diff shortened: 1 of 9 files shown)


ahmedhamodi:
Ahhh Vivian you continue to impress 🎉  - video looks great. Will review code shortly. Let me also find you another task 😅

src/Components/Admin/EnvironmentEditor/environment.js
```diff
@@ -102,16 +112,21 @@ export default function EnvironmentEditor({
...
+                friendly_name: environment.friendly_name,
+                description: environment.description,
+                scene_ids: newSceneIds,
+                is_published: environment.is_published,
+                is_previewable: environment.is_previewable,
+            });
+            setEnvironment(response.data);
+        } catch (error) {
+            handleError(error);
+        }
```
aaronabraham311, line 129: Do you think it's better to have the try catch here or should we handle it in the endpoints instead?

aaronabraham311 (commented):
Looks awesome Vivian! One thing that looked a bit off is that the background of the modal disappears. Is this supposed to happen and is there a way around it?

src/Components/Admin/EnvironmentEditor/environment.js
aaronabraham311: I think it would actually make more sense to handle errors in the endpoints if possible and just pass the `handleError` function as well. This way, we don't need to remember to handle errors when implementing the frontend. Would reduce effort imo

vivianliu0:
> Looks awesome Vivian! One thing that looked a bit off is that the background of the modal disappears. Is this supposed to happen and is there a way around it?

Ah yea, this is because the ErrorBoundary renders the ErrorModal component by itself as the fallback component rather than on top of the children components. I'll look to see if theres a way around it, and if it isn't possible, I can make the Error Modal a full screen dialog instead.

aaronabraham311:
> I'll look to see if theres a way around it, and if it isn't possible, I can make the Error Modal a full screen dialog instead.

Don't worry about the full screen dialog for now. If it isn't possible, we can work around it at a later time. Don't want to block you!

src/Components/Admin/EnvironmentEditor/environment.js
vivianliu0: Yea I agree this would be cleaner, updated PR with this change.

aaronabraham311:
I believe you're missing a few more endpoint calls in `environment.js`

aaronabraham311 (approved):

ahmedhamodi (approved):
lgtm :)

PR #52, opened 2021-05-19
Implemented Delete Scene Functionality for Scene Cards

andstun:
Wrote several functions that allow you to delete a scene from an escape room

### DIFF
--- a/src/Components/Admin/EnvironmentEditor/environment.js
+++ b/src/Components/Admin/EnvironmentEditor/environment.js
@@ -9,8 +9,14 @@ import Navbar from "../navbar";
 import EnvironmentBar from "./environmentBar";
 import SceneCard from "./sceneCard";
 import SceneModal from "./sceneModal";
+import DeleteModal from "../common/deleteModal";
 import { getScenario, editScenario } from "../../../lib/scenarioEndpoints";
-import { getScene, createScene, editScene } from "../../../lib/sceneEndpoints";
+import {
+    getScene,
+    createScene,
+    editScene,
+    deleteScene,
+} from "../../../lib/sceneEndpoints";
 import { useErrorHandler } from "react-error-boundary";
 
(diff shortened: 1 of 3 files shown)


ahmedhamodi:
Will review when the associated backend fix PR is merged (see https://github.com/uwblueprint/cdc-backend/pull/101)

src/Components/Admin/EnvironmentEditor/environment.js
```diff
@@ -180,6 +188,33 @@ export default function EnvironmentEditor({
...
+    const onDeleteModalSubmit = async () => {
+        await deleteScene(deleteSceneId, handleError);
+
+        const modifiedScenes = scenes.filter(
+            (scene) => scene.id !== deleteSceneId
+        );
+
+        const newEnvData = environment;
+        newEnvData.scene_ids = modifiedScenes.map((scene) => scene.id);
+        const newEnv = await editScenario(newEnvData, handleError);
```
aaronabraham311, line 210: @ahmedhamodi If you look at this logic here, on delete, we actually update the `scene_ids` field on the scenario. Are you planning on changing this so that the `deleteScene` endpoint automatically does this?

aaronabraham311 (commented):
Solid code, Kevin. Great work.

Also, please change your PR name to something that describes your changes.  A decent PR title would be 'Delete scene functionality'. Right now, it looks like it's just the first commit message  :)

src/Components/Admin/EnvironmentEditor/environment.js
ahmedhamodi: Yeah I've already implemented that in the PR mentioned above. It goes through all the scenarios and removes the associated scene.

This is better to do from endpoint side because it can handle all the other scenarios impacted by this -- the frontend should not know this information.

ahmedhamodi (approved):
Backend PR merged, this is good to go!

PR #56, opened 2021-05-22
Logout functionality from navbar

ahmedhamodi:
I noticed when screen sharing with Amolik that his default flow was to click this button on the navbar that did nothing, so wanted to implement some logic so that it actually works as expected!

Loom video:
https://www.loom.com/share/75dbe1c66ca945898074dc80368abc03

### DIFF
--- a/src/Components/Admin/navbar.js
+++ b/src/Components/Admin/navbar.js
@@ -45,6 +45,15 @@ export default function Navbar({ home }) {
         history.push("/admin");
     };
 
+    async function handleLogout() {
+        handleClose();
+        history.push("/login");
+        auth.signOut();
+        await httpGet(
+            process.env.REACT_APP_ADMIN_BASE_ENDPOINT + "admin_logout"
+        );
+    }
+
     return (
         <div className={classes.root}>
             <AppBar position="fixed">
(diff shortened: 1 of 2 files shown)


src/Components/Admin/navbar.js
```diff
@@ -49,39 +54,56 @@ export default function Navbar({ home }) {
...
+                                    onClick={async () => {
+                                        handleClose();
+                                        history.push("/login");
+                                        auth.signOut();
+                                        await httpGet(
+                                            process.env
+                                                .REACT_APP_ADMIN_BASE_ENDPOINT +
+                                                "admin_logout"
+                                        );
+                                    }}
```
aaronabraham311, line 100: I would prefer to have this pulled into another function than make it inline with `onClick`

src/Components/Admin/navbar.js
```diff
@@ -49,39 +54,56 @@ export default function Navbar({ home }) {
...
-                            className={classes.menu}
-                        >
-                            <MenuItem onClick={handleClose}>
-                                Documentation
-                            </MenuItem>
-                            <MenuItem onClick={handleClose}>Settings</MenuItem>
-                            <MenuItem onClick={handleClose}>Logout</MenuItem>
-                        </Menu>
-                    </div>
+                    {user && (
```
aaronabraham311, line 57: I assume this would prevent the menu from even showing up if you're logged out?

aaronabraham311 (commented):
Couple questions, overall looks good

src/Components/Admin/navbar.js
ahmedhamodi: The menu bar loads but the right most icon doesn't load so you cannot instigate menu.

Reason for this -- if we have a name or logo present in the navbar, we would probably still want that to load on login page. But no functionality (home, that menu dropdown) should load.

aaronabraham311: @andstun Perfect example of conditional rendering! You will need it in an upcoming ticket

aaronabraham311 (approved):

PR #57, opened 2021-05-26
Added "is_publishable" and "is_previewable" environment fields to the room cards

andstun:
Pretty much what the title of the PR suggests is what I did. The functionality works for editing rooms and was decided to not include these fields for creating rooms. Upon Ahmed's suggestion, I also added an "expected solve time" field - below the checkboxes; currently its value is fixed.

Let me know if this loom recording is slow for you guys - for some reason the playback was slow for me. I can retake it.
https://www.loom.com/share/8368782620a04bb78f3d517ef5f3b965

### DIFF
--- a/src/Components/Admin/Rooms/roomModal.js
+++ b/src/Components/Admin/Rooms/roomModal.js
@@ -36,14 +36,12 @@ export default function RoomModal({
     const [isPreviewable, setIsPreviewable] = React.useState(false);
 
     useEffect(() => {
-        if (room) {
-            setRoomName(room.name);
-            setFriendlyName(room.friendly_name);
-            setRoomDescription(room.description);
-            setIsPublished(room.is_published);
-            setIsPreviewable(room.is_previewable);
-            setRoomSolveTime(room.expected_solve_time);
-        }
+        setRoomName(room ? room.name : "");
+        setFriendlyName(room ? room.friendly_name : "");
+        setRoomDescription(room ? room.description : "");
+        setIsPublished(room ? room.is_published : false);
+        setIsPreviewable(room ? room.is_previewable : false);
+        setRoomSolveTime(room ? room.expected_solve_time : "");
     }, [room]);
 
     const handleRoomNameChange = (event) => {
@@ -62,16 +60,6 @@ export default function RoomModal({
         setRoomSolveTime(event.target.value);
     };
 
-    const handleModalCloseClick = () => {
-        handleModalClose();
-        setRoomName("");
-        setRoomDescription("");
-        setFriendlyName("");
-        setRoomSolveTime("");
-        setIsPublished(false);
-        setIsPreviewable(false);
-    };
-
     const handleModalSubmitClick = () => {
         if (isEdit) {
             handleSubmit({
@@ -82,9 +70,6 @@ export default function RoomModal({
                 is_previewable: isPreviewable,
                 expected_solve_time: roomSolveTime,
             });
-            setRoomSolveTime("");
-            setIsPublished(false);
-            setIsPreviewable(false);
         } else {
             handleSubmit({
                 name: roomName,
@@ -92,9 +77,6 @@ export default function RoomModal({
                 friendly_name: friendlyName,
             });
         }
-        setRoomName("");
-        setRoomDescription("");
-        setFriendlyName("");
     };
 
     const handleIsPublishedClick = () => {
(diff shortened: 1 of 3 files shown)


ahmedhamodi:
Nit: I think expected solve time should be with the other string fields, so above the checkboxes. Should be a quick change

aaronabraham311 (commented):
What is the expected behaviour when you are adding a room card? If I remember from Kevin, `api_schemas.py` doesn't allow you to define `is_published` and `is_previewable` when sending a POST request.

src/Components/Admin/admin.js
```diff
@@ -200,8 +202,9 @@ export default function Admin() {
                 friendly_name,
                 description,
                 scene_ids: editRoom.scene_ids,
-                is_published: editRoom.is_published,
-                is_previewable: editRoom.is_previewable,
+                is_published,
+                is_previewable,
+                expected_solve_time: editRoom.expected_solve_time,
```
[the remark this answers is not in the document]
ahmedhamodi: The PUT request allows for these fields, so I assume editScenario maps to PUT?

And yeah, this does not change expected solve time. You want to use the value put into that field you added in the modal popup!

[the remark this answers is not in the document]
aaronabraham311: But when we create a scene, are we going to prevent a user from setting the `is_published` and `is_previewable` parts of the room?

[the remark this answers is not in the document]
ahmedhamodi: Yeah those have default values, we should not let them set it

src/Components/Admin/Rooms/roomModal.js
```diff
@@ -140,38 +149,38 @@ export default function RoomModal({
...
-                    <Typography component="div" variant="h5">
-                        Expected Solve Time:{" "}
-                    </Typography>
-                    <TextField
-                        value={roomSolveTime}
-                        onChange={handleRoomSolveTimeChange}
-                        className={classes.textField}
-                    />
-                </div>
+                {isEdit ? (
```
aaronabraham311, line 152: Close! Use `isEdit && (<Component />)` instead of `isEdit ? (<Component />) : null`. The second version is used if you want to render another set of components if the condition is false.

src/Components/Admin/Rooms/roomModal.js
```diff
@@ -50,24 +58,61 @@ export default function RoomModal({
...
         setRoomDescription("");
         setFriendlyName("");
     };
 
+    const handleIsPublishedClick = () => {
+        if (isPublished === true) {
+            setIsPublished(false);
+        } else {
+            setIsPublished(true);
+        }
```
jaydhulia, line 105: Can be simplified to:
```
setIsPublished(!isPublished)
```
😄

src/Components/Admin/Rooms/roomModal.js
ahmedhamodi: Ah good point. Sorry for not mentioning that when we synced Kevin.

Personally, I still use the `<condition> ? <true logic> : null` in case we do need to add false logic in the future, but either is fine.

src/Components/Admin/Rooms/roomModal.js
andstun: I knew there must have been a smarter way to do that. Thank you for your wisdom!

src/Components/Admin/Rooms/roomModal.js
```diff
@@ -50,24 +58,53 @@ export default function RoomModal({
...
+    };
+
     const handleModalCloseClick = () => {
         handleModalClose();
         setRoomName("");
         setRoomDescription("");
         setFriendlyName("");
+        setRoomSolveTime("");
+        setIsPublished(false);
+        setIsPreviewable(false);
```
ahmedhamodi, line 72: I know Aaron initially added the logic here but I don't think it's needed? On each modal popup, we reset the fields in the useEffect on lines 38-47.

So in that useEffect, we can get rid of the if statement and do a ternary operator on each field. `room ? room.<field> : ""`. And we don't need to do this for edit only fields (room solve time, ispublished, and ispreviewable) since this flow is only for creating. On edit, the useEffect will reset the fields to the ones in the room's value. On create, it'll reset the exposed fields (just name, desc, friendly name) to empty strings.

I'll play around with this myself and push up a change, don't want to do delay this a few days to get this logic reworking done.

andstun: Okay, thanks.

ahmedhamodi (approved):
looks good to me now :) tested locally

PR #58, opened 2021-05-26
Navigation to inspector from react frontend

amoliksingh:


src/Components/Admin/EnvironmentEditor/sceneCard.js
```diff
@@ -87,15 +87,32 @@ export default function SceneCard({
...
+                                href={
+                                    process.env.REACT_APP_ADMIN_BACKEND_URL +
+                                    "/admin/scene/" +
+                                    scene.id
+                                }
+                                target="_blank"
+                                rel="noreferrer"
+                                style={{
+                                    textDecoration: "none",
+                                    color: "#000",
```
ahmedhamodi, line 103: I would've thought textDecoration resets color too? But I assume you tried without this so it's okay to keep.

ahmedhamodi (approved):
lgtm - small comment

src/Components/Admin/EnvironmentEditor/sceneCard.js
amoliksingh: Yea tried without the color attribute, it was still purple, hence it being here

PR #60, opened 2021-05-28
Fix image load time

ahmedhamodi:
Fixed it by converting large image from SVG to PNG.

Loom video:
https://www.loom.com/share/96d5e20fca344b1faac4983b76520a5d

Image showing load time:
image


aaronabraham311:
Did this conversion actually work? Could you perhaps add a latency screenshot from the networks panel? I was guessing that this is probably the issue lol

ahmedhamodi:
> Did this conversion actually work? Could you perhaps add a latency screenshot from the networks panel? I was guessing that this is probably the issue lol

Sure will do. It was 2ms when I tested after this change

aaronabraham311 (approved):

PR #61, opened 2021-05-29
Changes for scene redirect to aframe-inspector

amoliksingh:
Tested opening Living Room and Classroom, object modification worked as expected


ahmedhamodi (changes requested):
Let's also change the "handleCardClick" function on lines 54-60 so that there is no scene/{id} route (since that is just the inspector now)

ahmedhamodi (approved):
Works locally!

PR #63, opened 2021-06-02
View transitions work

ahmedhamodi:
Video demo:
https://www.loom.com/share/72ce76ba1e6f46b491d959dfdea51ea7

What's working in this PR?
- ability to drag transition around with scene in scenario editor
- ability to click "edit transitions" and a modal shows up (using it's own variables)

Todos after this PR?
- actually fetch relevant data on the "edit transitions" click (it's using scene data and fields)
- actually edit transition data upon submitting the popup form

### DIFF
--- a/src/Components/Admin/EnvironmentEditor/environment.js
+++ b/src/Components/Admin/EnvironmentEditor/environment.js
@@ -3,11 +3,12 @@ import { makeStyles } from "@material-ui/core/styles";
 import Button from "@material-ui/core/Button";
 import AccountBalanceIcon from "@material-ui/icons/AccountBalance";
 import AddIcon from "@material-ui/icons/Add";
-import { DragDropContext, Droppable } from "react-beautiful-dnd";
+import { DragDropContext, Droppable, Draggable } from "react-beautiful-dnd";
 
 import Navbar from "../navbar";
 import EnvironmentBar from "./environmentBar";
 import SceneCard from "./sceneCard";
+import TransitionCard from "./transitionCard";
 import SceneModal from "./sceneModal";
 import DeleteModal from "../common/deleteModal";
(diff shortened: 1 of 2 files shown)


src/Components/Admin/EnvironmentEditor/environment.js
```diff
@@ -159,11 +167,24 @@ export default function EnvironmentEditor({
         setEditModalOpen(true);
     };
 
+    const onTransitionEditClick = (sceneId) => {
+        const scene = scenes.find((scene) => scene.id === sceneId);
+
+        // TODO: get transition information (have to go up to scenario and match index from scene table)
+        setEditTransitionInfo(scene);
+        setEditTransitionModalOpen(true);
```
aaronabraham311, line 175: If it's this much work, do you think a restructuring of our table for this is necessary? I don't really know, hence why I am asking

aaronabraham311 (commented):
Is 'editing' the transition meta-pane actually just adding and moving around the individual transition pane? I believe in the ticket we had a nested structure. Is that what you will do next?

I don't really think having a form is necessary since we have the nested structure. Why did we not go with the rough design proposed in the ticket?

ahmedhamodi:
> Is 'editing' the transition meta-pane actually just adding and moving around the individual transition pane? I believe in the ticket we had a nested structure. Is that what you will do next?
> 
> I don't really think having a form is necessary since we have the nested structure. Why did we not go with the rough design proposed in the ticket?

Yeah this is right. I think I had told you initially that having that dynamic component popup would be too difficult, so I wanted to do a popup. Right now it's a form because it's in line with the edit scene popup, but I would change that to the view in the ticket (ie. each transition is a further square that can be moved around, and that will trigger an edit call).

It's a bit of work in itself, so thought a different PR for it made sense. Let me know.

aaronabraham311 (approved):
gif

PR #64, opened 2021-06-02
Added Regex Checking to Handle Incorrect Form Entries for Scenarios on Frontend 

andstun:
Room Description and Expected Solve Time are still bugged out: the Regex checking seems to fail regardless of any character, but doesn't trigger for the current values.

EDIT: Not sure exactly what the difference is between these two Regex expressions, but substituting `/^[?!.,a-zA-Z0-9 _-]{1,2000}$/` for `/^[?!.,a-zA-Z0-9 _-]{,2000}$/` in line 94 fixed the issue.


ahmedhamodi (commented):
According to offline discussion with Kevin, the description bug is no longer happening so I will test this tomorrow and approve if it works as expected. Good work :)

ahmedhamodi (approved):
Code looks good! Tested locally.

PR #65, opened 2021-06-03
Transitions popup (for editing order)

ahmedhamodi:
Video showing the feature:
https://www.loom.com/share/3770fa6539264786a1012055677377ca

Reworked the popup styling, see new image below:
image

### DIFF
--- a/src/Components/Admin/EnvironmentEditor/environment.js
+++ b/src/Components/Admin/EnvironmentEditor/environment.js
@@ -4,12 +4,14 @@ import Button from "@material-ui/core/Button";
 import AccountBalanceIcon from "@material-ui/icons/AccountBalance";
 import AddIcon from "@material-ui/icons/Add";
 import { DragDropContext, Droppable, Draggable } from "react-beautiful-dnd";
+import { useErrorHandler } from "react-error-boundary";
 
 import Navbar from "../navbar";
 import EnvironmentBar from "./environmentBar";
 import SceneCard from "./sceneCard";
-import TransitionCard from "./transitionCard";
 import SceneModal from "./sceneModal";
+import TransitionCard from "./transitionCard";
+import TransitionModal from "./transitionModal";
 import DeleteModal from "../common/deleteModal";
 import { getScenario, editScenario } from "../../../lib/scenarioEndpoints";
 import {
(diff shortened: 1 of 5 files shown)


src/Components/Admin/EnvironmentEditor/environment.js
```diff
@@ -214,10 +216,13 @@ export default function EnvironmentEditor({
...
-    const onTransitionModalSubmit = async () => {
+    const onTransitionModalSubmit = async (transitions) => {
         setEditTransitionModalOpen(false);
 
-        // TODO: edit transition data upon edit
+        const envData = environment;
+        envData.transitions.data = transitions;
+        const response = await editScenario(envData, handleError);
+        setEnvironment(response.data);
     };
```
aaronabraham311, line 226: Are we firing a request each time we change the transition order?

src/Components/Admin/EnvironmentEditor/transitionModal.js
```diff
@@ -0,0 +1,87 @@
...
+    return (
+        <Dialog open={modalOpen} onClose={handleModalClose}>
+            <DialogTitle>Edit Transitions Order</DialogTitle>
+            <DialogContent>
+                {transitions.map((transition, index) => {
+                    return (
+                        <div key={transition.id}>
+                            <h4>{transition.text}</h4>
+                            <h5>
+                                Transition {index + 1} of {transitions.length}
```
aaronabraham311, line 64: Could we reverse the order

aaronabraham311 (commented):
Nice work! Couple of comments

src/Components/Admin/EnvironmentEditor/environment.js
ahmedhamodi: When you "submit" the modal, then yes. I can add a quick check for if it's the same order?

src/Components/Admin/EnvironmentEditor/transitionModal.js
ahmedhamodi: What do you mean? Want transition 1 on bottom? I feel like the current order makes more sense

aaronabraham311: No I mean Transition {index + 1} of {transitions.length} above the actual text

ahmedhamodi: Oh I see. will do

aaronabraham311 (approved):

PR #67, opened 2021-06-05
Asset Cards for Admin Page

vivianliu0:
Loom Video:
https://www.loom.com/share/3c065d999f664709ba1bbf51aca48e9f


ahmedhamodi:
Yes, that over-flow issue is known and there is already a ticket for that (assigned to Kevin). It was there before, so agreed you don't need to resolve it here.

You have a small lint error if you can fix that up :)

ahmedhamodi (commented):
Code looks good.

Let's delete itemCard.js since it's not longer being used anywhere.

ahmedhamodi (approved):

PR #72, opened 2021-06-11
Model Viewer Integration

vivianliu0:
Loom Demo:
https://www.loom.com/share/f2dd03fe841e45519e49a09a7595bef1


src/Components/Admin/AssetModelViewer/assetModelViewer.js
```diff
@@ -0,0 +1,143 @@
...
+                        color="primary"
+                    >
+                        Save
+                    </Button>
+                </form>
+            </Drawer>
+            <main className={classes.content}>
+                <div className={classes.toolbar} />
+                <model-viewer
+                    src="Astronaut.gltf"
```
ahmedhamodi, line 132: I feel like this will check for Astronaut in the `public/` folder, not in the place the file actually exists. Might want to try importing the Astronaut file at the top of this file and then using that variable? Not sure if this is even the issue you are facing but just wanted to comment and let you know.

vivianliu0: Thanks yea, that was actually originally what was causing a problem, I was able to get the glb model showing up in the model viewer

ahmedhamodi: Exciting stuff. rest of the code in this PR looked good. Will review and test more closely when you push the working version :)

src/Components/Admin/AssetModelViewer/assetModelViewer.js
```diff
@@ -0,0 +1,153 @@
...
+    const [objectType, setObjectType] = useState("");
+    const [buttonText, setButtonText] = useState("Save");
+    const [modelGLB, setModelGLB] = useState("");
+
+    useEffect(() => {
+        const getObjectAsset = async () => {
+            const data = await getAsset(assetId, handleError);
+            setName(data.name);
+            setObjectType(data.obj_type);
+            // TODO: Replace with asset's s3_link once it contains a legit link
```
jaydhulia, line 61: This currently contains the legit link right?
It should be `asset_prefix` + `data.s3_key`,
where `asset_prefix` == `http://localhost:8888/static/` in dev
and `asset_prefix` == `https://dcc-bp-public-dev.s3.us-east-2.amazonaws.com/` for deployment

src/Components/Admin/AssetModelViewer/assetModelViewer.js
```diff
@@ -0,0 +1,153 @@
...
+                    </Typography>
+                    <TextField
+                        id="outlined-helperText"
+                        label="Asset Name"
+                        value={name}
+                        onChange={handleNameChange}
+                        variant="outlined"
+                    />
+
+                    <TextField
```
jaydhulia, line 119: A `Select` with options here would be a better option here imo, getting admins to select between `Background` and `Object` rather than having to know the exact thing. Because if they input `Object` instead of `object`, it will also error.

aaronabraham311: ^^ Agree with this

src/Components/Admin/AssetModelViewer/assetModelViewer.js
```diff
@@ -0,0 +1,153 @@
...
+    };
+
+    async function updateAsset(event) {
+        event.preventDefault();
+        setButtonText("Saved!");
+        setTimeout(() => {
+            setButtonText("Save");
+        }, 2000);
+        editAsset(assetId, name, objectType, asset.s3_key, handleError);
+    }
```
aaronabraham311, line 86: @ahmedhamodi is the way that we did `setTimeout` best practice? I almost never use `setTimeout` for frontend work because it could be blocking but idk.

src/Components/Admin/AssetModelViewer/assetModelViewer.js
```diff
@@ -0,0 +1,153 @@
...
+export default function AssetModelViewer({
+    match: {
+        params: { assetId },
+    },
+}) {
+    const classes = useStyles();
+    const handleError = useErrorHandler();
+    const [asset, setAsset] = useState({});
+    const [name, setName] = useState("");
+    const [objectType, setObjectType] = useState("");
```
aaronabraham311, line 52: Rather than using a string for this state, let's try using an enum instead, since there's pre-specified options. Enums are much cleaner and easier to change later. Ahmed may have a differing opinion

src/Components/Admin/AssetModelViewer/assetModelViewer.js
```diff
@@ -0,0 +1,153 @@
...
+    const [buttonText, setButtonText] = useState("Save");
+    const [modelGLB, setModelGLB] = useState("");
```
aaronabraham311, line 54: What is this state for?

aaronabraham311 (commented):
Great work, Vivian! Couple of comments:

- I was under the impression that the S3 links are already passed from the object endpoints @jaydhulia can you comment on this? Perhaps I'm not updated on this
- What are those options on the right end of the model-viewer? Is this something important? Are we able to hide it?
- As Jay mentioned, having a select dropdown would be better. I will scope better error handling for another ticket

src/Components/Admin/AssetModelViewer/assetModelViewer.js
ahmedhamodi: I've never used setTimeout before. I also don't think it's needed here. Was there any reason you added it Vivian?

src/Components/Admin/AssetModelViewer/assetModelViewer.js
vivianliu0: This could just be replaced with a constant

vivianliu0:
> * What are those options on the right end of the model-viewer? Is this something important? Are we able to hide it?
Those icons are from a different screen recording application I have on my laptop lolll, so not actually part of the chrome window

jaydhulia:
@aaronabraham311 @vivianliu0 
> I was under the impression that the S3 links are already passed from the object endpoints @jaydhulia can you comment on this? Perhaps I'm not updated on this

Yup, the S3 / local links are passed from the backend to the frontend. So we should be able to have that working in this PR. I put in a comment https://github.com/uwblueprint/cdc-frontend/pull/72#discussion_r650428302 about the TODO, which addresses this.
Basically create a new `.env` variable that reflects `asset_prefix` and set it to `http://localhost:8888/static/` for dev.

src/Components/Admin/AssetModelViewer/assetModelViewer.js
vivianliu0: I added it so that when the user clicks on the save button, it changes to "saved!" for 2 seconds before returning back to the original "save" state. Wasn't sure if there was a better way to do this without setTimeout

aaronabraham311: We can keep this in for now but I remember someone telling me using setTimeout on the frontend is not a good idea. However, don't want this to block you from moving forward so we can keep this in until the designers have a better idea

src/Components/Admin/AssetModelViewer/assetModelViewer.js
vivianliu0: Updated PR with new environment variable

src/Components/Admin/AssetModelViewer/assetModelViewer.js
vivianliu0: Replaced textfield with selector

src/Components/Admin/AssetModelViewer/assetModelViewer.js
ahmedhamodi: Sorry maybe I should've commented to be fully clear, but we should use an enum here. We'll also want to change any other location in the code base using strings here, but we can do that in a different ticket.

src/Components/Admin/AssetModelViewer/objectTypes.ts
```diff
@@ -0,0 +1,4 @@
+export enum ObjectTypes {
```
ahmedhamodi, line 1: It looks like you added the enum here properly, let's use it in the code! When we do `setObjectType`, you'll need to add a switch statement to check for "object", "background" and then a default which can throw an error.

src/Components/Admin/AssetModelViewer/assetModelViewer.js
vivianliu0: Mm yea I was running into some issues when I tried changing the state to an enum, the selector field would overlay with the "Object Type", I can try looking into it some more

vivianliu0: Update: fixed now :)

src/Components/Admin/AssetModelViewer/assetModelViewer.js
```diff
@@ -57,15 +57,29 @@ export default function AssetModelViewer({
...
+                case ObjectTypes.OBJECT:
+                    setObjectType(ObjectTypes.OBJECT);
+                    break;
+                case ObjectTypes.BACKGROUND:
+                    setObjectType(ObjectTypes.BACKGROUND);
+                    break;
+                default:
+                    setObjectType(ObjectTypes.NONE);
+                    throw new Error(
+                        "Object type must be either object or background"
```
aaronabraham311, line 79: How do we handle this error? Technically we should never reach this, correct?

src/Components/Admin/AssetModelViewer/assetModelViewer.js
```diff
@@ -122,18 +124,27 @@ export default function AssetModelViewer({
...
                             onChange={handleObjectTypeChange}
                         >
-                            <MenuItem value="object">{ObjectTypes.OBJECT}</MenuItem>
-                            <MenuItem value="background">{ObjectTypes.BACKGROUND}</MenuItem>
+                            <MenuItem value="object">
+                                {ObjectTypes.OBJECT}
+                            </MenuItem>
+                            <MenuItem value="background">
+                                {ObjectTypes.BACKGROUND}
+                            </MenuItem>
```
aaronabraham311, line 173: I don't know if this is worth the change, but we could do `<MenuItem value={ObjectTypes.BACKGROUND}>...`. Definitely a nit and this shouldn't block you from merging (feel free to dismiss this feedback)

aaronabraham311 (approved):
Excellent work!

src/Components/Admin/AssetModelViewer/assetModelViewer.js
vivianliu0: Yea, we can only ever receive the values that are set in the selector drop down (object and background). Currently the error is not handled, but if need be I can add this in the future.

ahmedhamodi:
This PR is missing a back button to navigate back to admin dashboard. Can do in a separate ticket.

ahmedhamodi (approved):
Looks great to me :)

PR #73, opened 2021-06-12
Add transition functionality

ahmedhamodi:
Loom demo: https://www.loom.com/share/f26555de2afc4f62bda41442c7a4295e


src/Components/Admin/EnvironmentEditor/transitionModal.js
```diff
@@ -44,6 +59,14 @@ export default function TransitionModal({
         setTransitions([...reorderedList]);
     };
 
+    const addTransition = () => {
+        const newTransition = {
+            text: prompt("Enter the text for the transition: "),
+        };
+
+        setTransitions([...transitions, newTransition]);
+    };
```
aaronabraham311, line 68: Lol we'll definitely be changing this in the future but works for now I guess

aaronabraham311 (approved):

src/Components/Admin/EnvironmentEditor/transitionModal.js
ahmedhamodi: We need to be careful with popups within popups. Will need to explore a better solution, but I'll wait until designs are complete before using the engineering time.

PR #77, opened 2021-06-15
Added password strength checker on admin signup flow

andstun:
Password restrictions:
- minimum 10 characters
- 1 lowercase char
- 1 uppercase char
- 1 symbol
- 1 number

https://www.loom.com/share/c4dca9628b304a2e845c1721b4a20811

### DIFF
--- a/src/Components/Login/signup.js
+++ b/src/Components/Login/signup.js
@@ -52,6 +52,15 @@ export default function Signup() {
         const errors = {};
 
         try {
+            const reg = new RegExp(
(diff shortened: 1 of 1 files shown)


ahmedhamodi (commented):
Let's override the error message when a weak password is given. Look at line 79 in `signup.js`. Let's list the actual requirement. Something like:

"Password is too weak. Requires a lower case letter, an upper case letter, a number, a symbol, and must be at least 10 characters in length."

ahmedhamodi (approved):

PR #78, opened 2021-06-16
Added error handling when exceeding character limits in scenario/scene creation flow

andstun:
The current flow only displays the character limit error message if all the characters are valid. If there is an invalid character, it will first display the invalid character error message. @ahmedhamodi let me know if this functionality should be any different.

https://www.loom.com/share/1513d40fa9194e9f916575a22a767d68


ahmedhamodi:
Yeah the order of error displaying is good imo. Will review code + test now

ahmedhamodi (approved):
Tested. lgtm

PR #79, opened 2021-06-16
Adding text-pane CRUD endpoint

amoliksingh:
here is a demo, https://www.loom.com/share/311c4825cbcf4665ab53869d3b1d6c5f

### DIFF
--- a/src/Components/Admin/ObjectEditor/object.js
+++ b/src/Components/Admin/ObjectEditor/object.js
@@ -164,7 +164,7 @@ export default function ObjectEditor({
     const deleteTransition = (index) => {
         const tempTexts = JSON.parse(JSON.stringify(texts));
         tempTexts.splice(index, 1);
-        for (var i = 0; i < tempTexts.length; i++) {
+        for (let i = 0; i < tempTexts.length; i++) {
             tempTexts[i].index = i;
         }
         setTexts(tempTexts);
--- a/src/Components/Admin/ObjectEditor/object.js
+++ b/src/Components/Admin/ObjectEditor/object.js
@@ -1,8 +1,6 @@
 import React, { useEffect, useState } from "react";
 import { makeStyles } from "@material-ui/core/styles";
-import AccountBalanceIcon from "@material-ui/icons/AccountBalance";
 import AddIcon from "@material-ui/icons/Add";
-import { DragDropContext, Droppable, Draggable } from "react-beautiful-dnd";
 import { useErrorHandler } from "react-error-boundary";
 import Select from "react-select";
 import DialogTitle from "@material-ui/core/DialogTitle";
@@ -16,8 +14,6 @@ import {
     KeyboardArrowUp,
 } from "@material-ui/icons";
 
-import Navbar from "../navbar";
-import DeleteModal from "../common/deleteModal";
 import { getPuzzle, editPuzzle } from "../../../lib/puzzleEndpoints";
 
 const useStyles = makeStyles((theme) => ({
@@ -101,11 +97,11 @@ export default function ObjectEditor({
             setPuzzleBody(data.animations_json.blackboardData);
             setPuzzleType(puzzleBody.componentType);
             setIsInteractable(data.is_interactable);
-            console.log(puzzleBody);
+            // console.log(puzzleBody);
         };
 
         getPuzzleBody();
-    }, [sceneId, objectId, handleError]);
+    }, [sceneId, objectId, puzzleBody, handleError]);
 
     const selectPuzzleType = (obj) => {
         if (obj) {
@@ -113,7 +109,7 @@ export default function ObjectEditor({
             data.componentType = obj.value;
             setPuzzleBody(data);
             setPuzzleType(obj.value);
-            console.log(puzzleBody);
+            // console.log(puzzleBody);
         }
     };
 
@@ -164,9 +160,9 @@ export default function ObjectEditor({
         setTransitions(tempTransitions);
     };
 
-    const handleSubmit = (obj) => {
+    const handleSubmit = () => {
         const savePuzzle = async () => {
-            const data = await editPuzzle(
+            await editPuzzle(
                 sceneId,
                 objectId,
                 isInteractable,
@@ -178,25 +174,25 @@ export default function ObjectEditor({
         savePuzzle();
     };
 
-    const handleRemove = (text) => {
-        const newBody = puzzleBody;
-        const newList = newBody.jsonData.data.filter(
-            (item) => item.text !== text
-        );
-        newBody.jsonData.data = newList;
-        setPuzzleBody(newBody);
-    };
+    // const handleRemove = (text) => {
+    //     const newBody = puzzleBody;
+    //     const newList = newBody.jsonData.data.filter(
+    //         (item) => item.text !== text
+    //     );
+    //     newBody.jsonData.data = newList;
+    //     setPuzzleBody(newBody);
+    // };
 
-    const handleAdd = (text) => {
-        const newBody = puzzleBody;
-        if (!newBody.jsonData.data) {
-            newBody.jsonData.data = [{ text: text }];
-        } else {
-            const newList = newBody.jsonData.data.concat({ text: text });
-            newBody.jsonData.data = newList;
-        }
-        setPuzzleBody(newBody);
-    };
+    // const handleAdd = (text) => {
+    //     const newBody = puzzleBody;
+    //     if (!newBody.jsonData.data) {
+    //         newBody.jsonData.data = [{ text: text }];
+    //     } else {
+    //         const newList = newBody.jsonData.data.concat({ text: text });
+    //         newBody.jsonData.data = newList;
+    //     }
+    //     setPuzzleBody(newBody);
+    // };
 
     return (
         <div className={classes.container}>
--- a/src/Components/Admin/ObjectEditor/object.js
+++ b/src/Components/Admin/ObjectEditor/object.js
@@ -68,7 +68,7 @@ export default function ObjectEditor({
     const classes = useStyles();
     const handleError = useErrorHandler();
     const [puzzleBody, setPuzzleBody] = useState({});
-    const [puzzleType, setPuzzleType] = useState("ordered-puzzle");
+    const [puzzleType, setPuzzleType] = useState("");
     const [animationsJson, setAnimationsJson] = useState({});
     const [isInteractable, setIsInteractable] = useState(false);
     const [texts, setTexts] = React.useState([]);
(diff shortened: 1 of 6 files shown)


src/lib/puzzleEndpoints.js
```diff
@@ -28,8 +28,12 @@ export const editPuzzle = async (
                 },
             }
         );
+        // console.log("response");
+        // console.log(response);
         return response;
     } catch (error) {
+        // console.log("error");
+        // console.log(error);
```
ahmedhamodi: Let's remove these console logs

ahmedhamodi:
Lint + merge conflict issues

src/Components/Admin/ObjectEditor/object.js
```diff
@@ -0,0 +1,263 @@
...
+                    )}
+                    options={puzzleTypeList}
+                    placeholder="Select puzzle type..."
+                    noResultsText="No puzzle types found"
+                    searchable={true}
+                    onChange={selectPuzzleType}
+                />
+            ) : null}
+            {isInteractable && puzzleType === "text-pane" ? (
+                <div>
```
jaydhulia, line 209: Might want to put this into a method and return this from there, and just call that method here. When we add the other puzzle types, this render will get super confusing if it is one giant return statement.

src/Components/Admin/ObjectEditor/object.js
```diff
@@ -0,0 +1,263 @@
...
+        }
+
+        const reorderedList = reorder(texts, sourceIndex, destinationIndex);
+
+        setTexts([...reorderedList]);
+    };
+
+    const addTransition = () => {
+        const newTransition = {
+            text: prompt("Enter the text for the transition: "),
```
jaydhulia, line 149: These aren't transitions, these are text popups

src/Components/Admin/ObjectEditor/object.js
```diff
@@ -0,0 +1,263 @@
...
+    const classes = useStyles();
+    const handleError = useErrorHandler();
+    const [puzzleBody, setPuzzleBody] = useState({});
+    const [puzzleType, setPuzzleType] = useState("");
+    const [animationsJson, setAnimationsJson] = useState({});
+    const [isInteractable, setIsInteractable] = useState(false);
+    const [texts, setTexts] = React.useState([]);
+
+    const puzzleTypeList = [
+        { value: "text-pane", label: "text-pane" },
```
jaydhulia, line 77: Can we update the actual text that is shown to the user to be more human-readable? `text-pane` doesn't sound very friendly lol, same for stuff like `rotation-controls`

src/Components/Admin/ObjectEditor/object.js
```diff
@@ -0,0 +1,263 @@
...
+                handleError
+            );
+        };
+
+        savePuzzle();
+    };
+
+    return (
+        <div className={classes.container}>
+            {isInteractable ? (
```
jaydhulia, line 196: Missing functionality: toggling whether an object is `isInteractable` or not should also be present (it was present in the original inspector PR, I think it didn't get migrated here?)

src/Components/Admin/ObjectEditor/object.js
```diff
@@ -0,0 +1,263 @@
...
+                { sceneId, objectId, isInteractable, animationsJson },
+                handleError
+            );
+        };
+
+        savePuzzle();
+    };
+
+    return (
+        <div className={classes.container}>
```
jaydhulia, line 195: Not an UI person, but I think the iFrame could use some padding on the left lol

jaydhulia:
Made some comments about things I observed, @ahmedhamodi or @aaronabraham311 probably better people to review it from React perspective

src/Components/Admin/ObjectEditor/object.js
```diff
@@ -0,0 +1,318 @@
...
+    button: {
+        display: "flex",
+        justifyContent: "center",
+        padding: "16px 12px",
+        backgroundColor: "#E2E5ED",
+        borderRadius: "12px",
+    },
+}));
+
+class TextPaneView extends React.Component {
```
aaronabraham311, line 58: Why are we using class components? We should be using functional components to align with the rest of the code base

src/Components/Admin/ObjectEditor/object.js
```diff
@@ -0,0 +1,318 @@
...
+    constructor(props) {
+        super(props);
+        this.state = {
+            texts: this.props.texts,
+        };
+        const tempTexts = this.props.texts;
+        for (let i = 0; i < tempTexts.length; i++) {
+            tempTexts[i]["index"] = i;
+        }
```
aaronabraham311, line 67: Let's try to avoid the for(let...) loops in JS if possible. Can we convert this into a JS function using `map`?

src/Components/Admin/ObjectEditor/object.js
aaronabraham311: Could we also move this component into another file? Would be bit easier to maintain

src/Components/Admin/ObjectEditor/object.js
```diff
@@ -0,0 +1,318 @@
...
+    const [isInteractable, setIsInteractable] = useState(false);
+
+    const puzzleTypeList = [
+        { value: "text-pane", label: "Text Puzzle" },
+        { value: "rotation-controls", label: "Rotation Puzzle" },
+        { value: "keypad", label: "Keypad Puzzle" },
+        { value: "visual-pane", label: "Visual Puzzle" },
+        { value: "jigsaw-puzzle", label: "Jigsaw Puzzle" },
+        { value: "ordered-puzzle", label: "Ordered Puzzle" },
+    ];
```
aaronabraham311, line 214: Is it possible to convert this into some sort of enum or another constant variable?

src/Components/Admin/ObjectEditor/object.js
```diff
@@ -0,0 +1,318 @@
...
+
+    useEffect(() => {
+        const getPuzzleBody = async () => {
+            const data = await getPuzzle(sceneId, objectId, handleError);
+            setAnimationsJson(data.animations_json);
+            if (JSON.stringify(data.animations_json) === JSON.stringify({})) {
```
aaronabraham311, line 220: Oh interesting way of doing null check. Why do we need to do `JSON.stringify` to check if it's an empty object?

src/Components/Admin/ObjectEditor/object.js
```diff
@@ -0,0 +1,318 @@
...
+                setPuzzleType(
+                    data.animations_json.blackboardData.componentType
+                );
+                setIsInteractable(data.is_interactable);
+                const puzBody = data.animations_json.blackboardData;
+                if (puzBody.componentType === "text-pane") {
+                    const tempTexts = puzBody.jsonData.data;
+                    for (let i = 0; i < puzBody.jsonData.data.length; i++) {
+                        tempTexts[i]["index"] = i;
+                    }
```
aaronabraham311, line 234: See previous comment about for loops

src/Components/Admin/ObjectEditor/object.js
```diff
@@ -0,0 +1,318 @@
...
+        }
+    };
+
+    const saveTexts = (texts) => {
+        const animCopy = animationsJson;
+        const puzzleBodyCopy = puzzleBody;
+        const textsCopy = texts;
+        for (let i = 0; i < textsCopy.length; i++) {
+            delete textsCopy["index"];
+        }
```
aaronabraham311, line 265: Can you explain this logic? I have never seen the usage of `delete` in JS lol

aaronabraham311 (commented):
Nice work Amolik! Couple of comments (mostly code style)

src/Components/Admin/ObjectEditor/object.js
amoliksingh: TextPaneView needs to call ObjectEditor's saveTexts function, is it possible to do this if it were a functional component, if so, how would I do it?

src/Components/Admin/ObjectEditor/object.js
amoliksingh: The index field is something I used to display the Text {index} of {texts.length} logic for the text cards, it shouldn't actually be a field sent to the backend when saving.

amoliksingh: Update: Reworked the logic to remove the usage of "index" as a key for texts

src/Components/Admin/ObjectEditor/object.js
amoliksingh: Switched to using Object.keys(puzzleBody).length === 0 instead

src/Components/Admin/ObjectEditor/object.js
aaronabraham311: I suggest making a wrapper class that wraps ObjectEditor and TextPaneView and create functions in the wrapper class that can be passed to both components.

src/lib/puzzleEndpoints.js
```diff
@@ -0,0 +1,35 @@
...
+) => {
+    try {
+        const response = await httpPut(
+            baseEndpoint + `scene/${sceneId}/object/${objectId}/puzzle`,
+            {
+                is_interactable: isInteractable,
+                animations_json: animationsJson,
+            }
+        );
+        alert("Saved puzzle CRUD changes for object with id: " + objectId);
```
ahmedhamodi, line 29: This should not be done in the puzzleEndpoints. Should be done in the function that calls "editPuzzle"

Other flows might call this that aren't explicit (to the user) that they are editing a puzzle

src/Components/Admin/ObjectEditor/object.js
```diff
@@ -0,0 +1,289 @@
...
+                    options={puzzleTypeList}
+                    placeholder="Select puzzle type..."
+                    noResultsText="No puzzle types found"
+                    searchable={true}
+                    onChange={selectPuzzleType}
+                />
+            ) : null}
+            {isInteractable && puzzleType === "text-pane" ? (
+                <TextPaneView
+                    saveTexts={saveTexts}
```
ahmedhamodi, line 282: To answer your question Amolik, this here will work even if TextPaneView is defined in a different file. You are passing the function, they don't need to be defined in same file.

Let's move TextPaneView to a new file.

src/Components/Admin/ObjectEditor/object.js
ahmedhamodi: Let's move to separate file. I'm okay with this being a class component - they work together so not a big deal.

In the future, let's do functional components in the cdc-frontend repo. I know aframe-inspector is outdated and still uses class components, so you can keep using class components there.

ahmedhamodi (commented):
Please do the proposed actions and then I'll be happy to review. Code looks good to me.

src/Components/Admin/ObjectEditor/object.js
amoliksingh: Tried to use enum, but got an error saying " 'enum' declarations can only be used in TypeScript files. "

ahmedhamodi (approved):
lgtm, works as expected locally. Styling needs work, but that is for later.

PR #83, opened 2021-06-26
Move save button to object.js

amoliksingh:
Tested changing object 1 to isInteractable as false, and different puzzle types.
Tested changing texts for object 4, with different ordering, adding, or deleting.
Tested setting isInteractable property to true for new objects and changing puzzle types.


ahmedhamodi (commented):
Code changes look good. Going to test locally before approving

ahmedhamodi (approved):

PR #85, opened 2021-06-26
Edit Asset

vivianliu0:
https://www.loom.com/share/46ea88caa61c4a469d5599e95ad02516

### DIFF
--- a/src/Components/Admin/AssetModelViewer/assetModelViewer.js
+++ b/src/Components/Admin/AssetModelViewer/assetModelViewer.js
@@ -133,7 +133,7 @@ export default function AssetModelViewer({
     }
 
     const isEmpty = (error) => {
-        return error == "";
+        return error === "";
     };
 
     return (
--- a/src/Components/Admin/AssetModelViewer/assetModelViewer.js
+++ b/src/Components/Admin/AssetModelViewer/assetModelViewer.js
@@ -15,6 +15,7 @@ import MenuItem from "@material-ui/core/MenuItem";
 import FormControl from "@material-ui/core/FormControl";
 import InputLabel from "@material-ui/core/InputLabel";
 import { ObjectTypes } from "./objectTypes.ts";
+import empty from "is-empty";
 
 const drawerWidth = 17;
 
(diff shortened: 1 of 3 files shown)


src/Components/Admin/AssetModelViewer/assetModelViewer.js
```diff
@@ -149,6 +162,8 @@ export default function AssetModelViewer({
                         value={name}
                         onChange={handleNameChange}
                         variant="outlined"
+                        error={!empty(nameError)}
```
aaronabraham311, line 165: Oh interesting why do we have `empty`? Couldn't we just compare against `""`? Or am I missing something?

aaronabraham311 (commented):
Think this is pretty good. @ahmedhamodi thoughts on the error message?

src/Components/Admin/AssetModelViewer/assetModelViewer.js
vivianliu0: Oh yup, we can do that instead. I just copied the validation logic from login.js which was using this for a slightly more complicated use case.

ahmedhamodi: We could, I think empty() is safer and also checks for null or other "empty"-type states. 

This covers the other states they detect.
https://www.npmjs.com/package/is-empty

ahmedhamodi (commented):
image

I'm able to fully remove the name and save. We should not allow this.

Note: actual save results in error, so definitely should not be allowing this:
image

ahmedhamodi:
> thoughts on the error message?

I think it's fine!

vivianliu0:
> I'm able to fully remove the name and save. We should not allow this.

Hmm weird, on my end I'm seeing that it disallows an empty field
image

ahmedhamodi:
> Hmm weird, on my end I'm seeing that it disallows an empty field

Oh I was on an older version of your branch for some reason o_o, I pulled and yeah I see that behaviour.

Okay, this lgtm then!

ahmedhamodi (approved):

src/Components/Admin/AssetModelViewer/assetModelViewer.js
```diff
@@ -15,6 +15,7 @@ import MenuItem from "@material-ui/core/MenuItem";
 import FormControl from "@material-ui/core/FormControl";
 import InputLabel from "@material-ui/core/InputLabel";
 import { ObjectTypes } from "./objectTypes.ts";
+import empty from "is-empty";
```
ahmedhamodi, line 18: You'll need to remove the unused import now too with the usage removed.

ahmedhamodi (approved):

PR #86, opened 2021-06-26
Clear blackboard data upon puzzle select

amoliksingh:
https://www.loom.com/share/063b55cb1d6248f980381a94db9608b9


ahmedhamodi (approved):

jaydhulia (changes requested):
Summarized the approach that we should follow on slack, please check that @amoliksingh

jaydhulia (approved):

ahmedhamodi:
According to Jay there is a small bug fix needed? Will wait for you to push that up Amolik to re-review

ahmedhamodi (approved):

PR #87, opened 2021-06-28
Navigate back to login on 403 forbidden error

ahmedhamodi:
Navigate back to login on 403 forbidden error + some css styling

Loom video: https://www.loom.com/share/0de7c1eee669402a9e4335e2cd0c0d76

### DIFF
--- a/src/Components/Admin/common/errorModal.js
+++ b/src/Components/Admin/common/errorModal.js
@@ -5,6 +5,8 @@ import DialogContent from "@material-ui/core/DialogContent";
 import DialogActions from "@material-ui/core/DialogActions";
 import DialogTitle from "@material-ui/core/DialogTitle";
 import { makeStyles } from "@material-ui/core/styles";
+import { auth } from "../../../firebaseCredentials";
+import { useHistory } from "react-router-dom";
 
 const useStyles = makeStyles(() => ({
     buttonContainer: {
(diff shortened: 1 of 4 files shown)


src/Components/Admin/common/errorModal.js
```diff
@@ -43,10 +47,23 @@ export default function ErrorModal({ error, resetErrorBoundary }) {
...
-                    onClick={resetErrorBoundary}
+                    onClick={() => {
+                        // TODO: send email to us
+                    }}
+                    className={classes.deleteButton}
+                >
+                    Report
+                </Button>
+                <Button
+                    onClick={() => {
```
jaydhulia, line 58: If I understand this code correctly, won't this log you out on _any_ error? So if you have a regex error and we show this error modal, if they hit close, this will log them out rather than just closing the modal.

src/Components/Admin/common/errorModal.js
```diff
@@ -43,10 +47,23 @@ export default function ErrorModal({ error, resetErrorBoundary }) {
...
+                        // signout, go to /login route
+                        auth.signOut();
+                        history.push("/login");
+                        resetErrorBoundary();
+                    }}
                     className={classes.deleteButton}
```
jaydhulia, line 64: `deleteButton`? This isn't the delete button though

jaydhulia (approved):

PR #88, opened 2021-06-30
PR for Visual pane

amoliksingh:
Loom demo: https://www.loom.com/share/2856796831a24c909bf1428c41f96cdf


src/Components/Admin/ObjectEditor/visualpaneview.js
```diff
@@ -0,0 +1,137 @@
...
+                </div>
+            ) : null}
+            {caption !== "" ? (
+                <div>
+                    Caption: {caption}
+                    <IconButton onClick={() => deleteCaption()}>
+                        <DeleteForever />
+                    </IconButton>
+                </div>
+            ) : null}
```
ahmedhamodi, line 134: The conditions here are just negations of each other, so instead of null put the other div and this should still function the same way.

jaydhulia (approved):
Looks good to me! I just have one scenario that I would like you to test: what happens if you save with no caption? Aka the caption being `""`. Is A-frame able to render this on the backend or does it error? If it errors, then we should enforce that caption isn't "" similar to how we did for text-pane.

amoliksingh:
> Looks good to me! I just have one scenario that I would like you to test: what happens if you save with no caption? Aka the caption being `""`. Is A-frame able to render this on the backend or does it error? If it errors, then we should enforce that caption isn't "" similar to how we did for text-pane.

Showed demo during meeting, A-frame is still able to render this, so it works as expected

PR #89, opened 2021-06-30
Auth consistency fixes

ahmedhamodi:
Changes:
- implemented a second field within the UserContext that is the field which indicates whether we are logged in with a valid admin user
- this additional field only gets recomputed when auth state changes
- if tokens are still set, no issues (admin user maintains access)
- if tokens are lost, this will fail and show a 403, which in a previous PR I added an okay button which will handle all state issues and move go /login route
- kept "user" as we can use that to pull fields like `displayName` and other info that firebase stores

What to test:
- logical analysis of code for any edge case where we have invalid state
- try to duplicate this when running locally
- try basic login / logout / clear cookies flows

Loom video:
https://www.loom.com/share/fb739486a1a14868a947048ccbdfd6f0

### DIFF
--- a/src/Components/Admin/admin.js
+++ b/src/Components/Admin/admin.js
@@ -114,7 +114,7 @@ const useStyles = makeStyles((theme) => ({
 
 export default function Admin() {
     const classes = useStyles();
-    const user = useContext(UserContext);
+    const { user } = useContext(UserContext);
     const handleError = useErrorHandler();
 
     const [value, setValue] = React.useState("rooms");
@@ -287,7 +287,7 @@ export default function Admin() {
                     Admin Dashboard 😎
                 </Typography>
                 <Typography component="div" variant="h6">
-                    Welcome {user.displayName}
+                    Welcome {user?.displayName}
                 </Typography>
 
                 <div className={classes.root}>
--- a/src/Components/Admin/navbar.js
+++ b/src/Components/Admin/navbar.js
@@ -29,7 +29,6 @@ const useStyles = makeStyles(() => ({
 export default function Navbar({ home }) {
     const classes = useStyles();
     const history = useHistory();
-    const user = useContext(UserContext);
     const [anchorEl, setAnchorEl] = React.useState(null);
     const open = Boolean(anchorEl);
 
@@ -47,12 +46,14 @@ export default function Navbar({ home }) {
 
     async function handleLogout() {
         handleClose();
-        history.push("/login");
         auth.signOut();
         await httpGet(
             process.env.REACT_APP_ADMIN_BASE_ENDPOINT + "admin_logout"
         );
+        history.push("/login");
     }
+
+    const { isAdmin } = useContext(UserContext);
 
     return (
         <div className={classes.root}>
@@ -63,7 +64,7 @@ export default function Navbar({ home }) {
                             <HomeIcon />
                         </IconButton>
                     )}
-                    {user && (
+                    {isAdmin && (
                         <div className={classes.profileEnd}>
                             <IconButton
                                 aria-label="account of current user"
--- a/src/Providers/UserProviders.jsx
+++ b/src/Providers/UserProviders.jsx
@@ -20,10 +20,11 @@ class UserProvider extends Component {
                 const user = await generateUserDocument(userAuth);
                 this.setState({ user });
             }
-            // TODO: in the future do an authentication route that returns name and email of user?
+
+            // validate cookies are set appropriately on this user instance
             try {
-                const isValidUser = await httpGet(baseEndpoint + "scenarios");
-                this.setState({ isAdmin: isValidUser !== undefined });
+                const response = await httpGet(baseEndpoint + "user_profile");
+                this.setState({ isAdmin: response.status === 200 });
             } catch {
                 this.setState({ isAdmin: false });
             }
--- a/src/Providers/UserProviders.jsx
+++ b/src/Providers/UserProviders.jsx
@@ -1,10 +1,14 @@
 import React, { Component, createContext } from "react";
 import { auth, generateUserDocument } from "../firebaseCredentials.js";
+import { httpGet } from "../lib/dataAccess";
+
+const baseEndpoint = process.env.REACT_APP_ADMIN_BASE_ENDPOINT;
 
 export const UserContext = createContext({ user: null });
 class UserProvider extends Component {
     state = {
         user: null,
+        isAdmin: false,
     };
 
     componentDidMount = async () => {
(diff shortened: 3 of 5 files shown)


src/Providers/UserProviders.jsx
```diff
@@ -16,12 +20,21 @@ class UserProvider extends Component {
                 const user = await generateUserDocument(userAuth);
                 this.setState({ user });
             }
+            // TODO: in the future do an authentication route that returns name and email of user?
```
ahmedhamodi, line 23: If we decide to implement this route in the future, it can also lead to the removal of `user` and the generateUserDocument logic.

However, it is not necessary.

src/Components/Admin/admin.js
```diff
@@ -114,7 +114,7 @@ const useStyles = makeStyles((theme) => ({
 
 export default function Admin() {
     const classes = useStyles();
-    const user = useContext(UserContext);
+    const { user } = useContext(UserContext);
```
aaronabraham311, line 117: Just curious, are we able to access the admin's Google Account picture? The designers have put this as a feature, but I wasn't sure if that's possible

src/Components/Admin/navbar.js
```diff
@@ -47,13 +46,15 @@ export default function Navbar({ home }) {
 
     async function handleLogout() {
         handleClose();
-        history.push("/login");
         auth.signOut();
         await httpGet(
             process.env.REACT_APP_ADMIN_BASE_ENDPOINT + "admin_logout"
         );
+        history.push("/login");
```
aaronabraham311, line 53: Why is this reorder needed?

src/Providers/UserProviders.jsx
```diff
@@ -1,10 +1,14 @@
...
+import { httpGet } from "../lib/dataAccess";
+
+const baseEndpoint = process.env.REACT_APP_ADMIN_BASE_ENDPOINT;
 
 export const UserContext = createContext({ user: null });
 class UserProvider extends Component {
     state = {
         user: null,
+        isAdmin: false,
     };
```
aaronabraham311, line 12: Any plan on refactoring this to functional components in the future?

aaronabraham311 (approved):
Codewise, I don't see any issues. Just a few questions. I couldn't run this locally because I am way too behind on the repo's changes, so I will leave local testing to @amoliksingh or @jaydhulia

src/Components/Admin/navbar.js
ahmedhamodi: Since the routing is now dependent on isAdmin, which gets recomputed on auth state changes, it is needed to be after the auth.signOut and the resetting of the cookies which the admin_logout does.

src/Providers/UserProviders.jsx
ahmedhamodi: Hmm, not high priority. Class + functional components work together as far as I know

src/Components/Admin/admin.js
ahmedhamodi: I haven't looked into it. Will probably need to do a different ticket for this.

My initial guess is no since you can make a firebase account with a non Google account, so that won't even be an accessible field. And specifically for DCC, they are using outlook emails so they definitely won't have a google account picture.

jaydhulia (changes requested):
Here are the issues I found:
- On the login page (and `/signup` page) itself, when I am not logged in, it makes a request to the backend to `/api/admin/v1/scenarios`, even though an admin is not logged in - this returns a 403 from BE, but this request shouldn't be made in the first place - Loom video: https://www.loom.com/share/4d4c9710134840a0ae10e18ddc494b65
- If an admin logs out, and goes to the `/admin` page (which they may have bookmarked), instead of being shown the login screen, it just shows a completely blank screen. Loom video: https://www.loom.com/share/b8afb9e0f03a49f3ac79d02b1d15b49c
- If an admin is logged in, and they go to the `/login` or `/signup` page, the login form shows up for a second and it just shows a completely blank screen, where you have no option other than to log out. Loom video (same as the second, later on in the video) - https://www.loom.com/share/b8afb9e0f03a49f3ac79d02b1d15b49c

ahmedhamodi:
> * On the login page (and `/signup` page) itself, when I am not logged in, it makes a request to the backend to `/api/admin/v1/scenarios`, even though an admin is not logged in - this returns a 403 from BE, but this request shouldn't be made in the first place

Ideally this makes a call to something like /api/admin/v1/authenticate or something similar. It's the logic we use to see if the cookies are set and are valid. For now it's a /scenarios call. What do you recommend is done instead? No call doesn't make much sense to me because we want to route them to /admin if they are in a valid logged in state when on /login.

> * If an admin logs out, and goes to the `/admin` page (which they may have bookmarked), instead of being shown the login screen, it just shows a completely blank screen.
> * If an admin is logged in, and they go to the `/login` or `/signup` page, the login form shows up for a second and it just shows a completely blank screen, where you have no option other than to log out.

Fixed both.

jaydhulia:
> Ideally this makes a call to something like /api/admin/v1/authenticate or something similar. It's the logic we use to see if the cookies are set and are valid. For now it's a /scenarios call. What do you recommend is done instead? No call doesn't make much sense to me because we want to route them to /admin if they are in a valid logged in state when on /login.

Ah that's what you are making it for. Cool, that's fine for now, we can update this to use a `user_profile` route once we add that to BE routes.

ahmedhamodi:
> Ah that's what you are making it for. Cool, that's fine for now, we can update this to use a `user_profile` route once we add that to BE routes.

Yup! Left a todo in the code for this already

src/Providers/UserProviders.jsx
```diff
@@ -16,12 +20,21 @@ class UserProvider extends Component {
                 const user = await generateUserDocument(userAuth);
                 this.setState({ user });
             }
+            // TODO: in the future do an authentication route that returns name and email of user?
+            try {
+                const isValidUser = await httpGet(baseEndpoint + "scenarios");
+                this.setState({ isAdmin: isValidUser !== undefined });
```
jaydhulia, line 26: Works for now, in the future we should ideally check if the response code was 403 or not. If it was something like 500, then that doesn't mean the user isn't logged in, just that the backend server is temporarily unavailable.

jaydhulia (approved):
Added a minor comment, but approving because that can be done in the future. LGTM otherwise.

src/Providers/UserProviders.jsx
ahmedhamodi: Hmm good point. Will make a mental note of this

jaydhulia: I added a PR on the backend, now we should be able to check 200 vs 403 response codes.

jaydhulia:
Backend PR up: https://github.com/uwblueprint/cdc-backend/pull/138, once that is merged, you can make a GET request to `/api/admin/v1/user_profile`

aaronabraham311 (approved):

jaydhulia (approved):

PR #90, opened 2021-07-01
Asset Delete

vivianliu0:
Loom Demo: https://www.loom.com/share/db1840fc26c34b909eef21d324c1425f

### DIFF
--- a/src/Components/Admin/admin.js
+++ b/src/Components/Admin/admin.js
@@ -27,7 +27,7 @@ import {
 } from "../../lib/scenarioEndpoints";
 import { UserContext } from "../../Providers/UserProviders";
 import { getAllScenes } from "../../lib/sceneEndpoints";
-import { getAllAssets, createAsset } from "../../lib/assetEndpoints";
+import { getAllAssets, createAsset, deleteAsset } from "../../lib/assetEndpoints";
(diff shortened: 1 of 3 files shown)


src/Components/Admin/admin.js
```diff
@@ -279,6 +287,25 @@ export default function Admin() {
...
+
+    const handleDeleteAssetCancel = () => {
+        setDeleteAssetId(null);
+        setDeleteAssetModalOpen(false);
+    };
+
+    const handleDeleteAssetSubmit = async () => {
+        await deleteAsset(deleteAssetId, handleError);
+
+        const modifiedAssets = assets.filter((env) => env.id !== deleteAssetId);
```
ahmedhamodi, line 303: Nit: the iterated object variable name in the filter shouldn't be `env`. Can we do `asset` instead?

ahmedhamodi (commented):
Code looks good. One minor nit

src/Components/Admin/admin.js
vivianliu0: Thanks, updated :)

ahmedhamodi (approved):
Tested locally, works as expected. Good work 🎉

PR #95, opened 2021-07-07
PR for adding header text

amoliksingh:
Changes for adding header text
Loom video: https://www.loom.com/share/e75a35cebd1a4b7d902aa1486cc072d8


src/Components/Admin/ObjectEditor/object.js
```diff
@@ -161,6 +179,24 @@ export default function ObjectEditor({
...
+    const addHeader = () => {
+        const newText = {
+            text: prompt("Enter the text for the header: "),
+        };
+
+        if (newText.text) {
+            if (newText.text.length <= 100) {
+                setHeader(newText.text);
+            } else {
+                alert("Error: Maximum header text length is 100 characters");
```
ahmedhamodi, line 191: When we move away from `prompt` for getting input from admin, we can also replace this `alert` with a proper error text below the input field in the popup. For now, this is fine.

ahmedhamodi (approved):
lgtm

PR #96, opened 2021-07-07
Redo UserContext + fix remaining auth bugs

ahmedhamodi:
Loom demo:
https://www.loom.com/share/f9ca258391b34d4b972b3a32ba0bd2ee

### DIFF
--- a/src/Components/Admin/navbar.js
+++ b/src/Components/Admin/navbar.js
@@ -61,7 +61,7 @@ export default function Navbar({ home }) {
         }
     }
 
-    const { isAdmin } = useContext(UserContext);
+    const { user } = useContext(UserContext);
 
     return (
         <div className={classes.root}>
@@ -72,7 +72,7 @@ export default function Navbar({ home }) {
                             <HomeIcon />
                         </IconButton>
                     )}
-                    {isAdmin && (
+                    {user && (
                         <div className={classes.profileEnd}>
                             <IconButton
                                 aria-label="account of current user"
--- a/src/Providers/UserProviders.jsx
+++ b/src/Providers/UserProviders.jsx
@@ -1,30 +1,25 @@
 import React, { Component, createContext } from "react";
-import { auth, generateUserDocument } from "../firebaseCredentials.js";
+import { auth } from "../firebaseCredentials.js";
 import { httpGet } from "../lib/dataAccess";
 
 const baseEndpoint = process.env.REACT_APP_ADMIN_BASE_ENDPOINT;
 
-export const UserContext = createContext({ user: null });
+export const UserContext = createContext({ isAdmin: false });
 class UserProvider extends Component {
     state = {
-        user: null,
         isAdmin: false,
+        displayName: "",
     };
 
     componentDidMount = async () => {
-        auth.onAuthStateChanged(async (userAuth) => {
-            if (!userAuth) {
-                this.setState({ user: null });
-            }
-            if (userAuth && userAuth.emailVerified) {
-                const user = await generateUserDocument(userAuth);
-                this.setState({ user });
-            }
-
+        auth.onAuthStateChanged(async () => {
             // validate cookies are set appropriately on this user instance
             try {
                 const response = await httpGet(baseEndpoint + "user_profile");
-                this.setState({ isAdmin: response.status === 200 });
+                this.setState({
+                    isAdmin: response.status === 200,
+                    displayName: response.data.display_name,
+                });
             } catch {
                 this.setState({ isAdmin: false });
             }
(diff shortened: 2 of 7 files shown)


src/Components/Admin/navbar.js
```diff
@@ -61,7 +61,7 @@ export default function Navbar({ home }) {
         }
     }
 
-    const { isAdmin } = useContext(UserContext);
+    const { user } = useContext(UserContext);
```
aaronabraham311, line 64: Can you explain why we changed this yet again? I remember we changed it to `isAdmin` in the previous auth PR

src/Routes.js
```diff
@@ -11,9 +11,15 @@ import AssetModelViewer from "./Components/Admin/AssetModelViewer/assetModelView
...
 export default function Routes() {
-    const { isAdmin } = useContext(UserContext);
+    const { isLoading, user } = useContext(UserContext);
 
-    return isAdmin ? (
+    useEffect(() => {
+        if (isLoading) {
+            return;
+        }
+    }, [isLoading]);
```
aaronabraham311, line 20: What does this do?

src/Providers/UserProviders.jsx
```diff
@@ -1,45 +1,99 @@
...
-        );
     }
-}
 
-export default UserProvider;
+    return (
+        <UserContext.Provider
+            value={{
+                isLoading: authState.state === AUTH_STATES.LOADING,
+                user: authState.user ?? null,
```
aaronabraham311, line 92: Learned something new today

aaronabraham311 (commented):
I don't see any problem points from a code perspective. I'll leave it to others to test this locally.

src/Routes.js
ahmedhamodi: I'll try removing it and see if it works but basically the idea is that if you are loading the routes without the user set (but in a loading state), recursively "wait" (ie. return nothing) until it is set, and then the rest of the routes file will load.

src/Components/Admin/navbar.js
ahmedhamodi: We did `isAdmin` since `user` was set (as a deprecated field that was no longer being used). Now that we removed the user document generation and the user object, I think user is a more logical name for this field.

src/Providers/UserProviders.jsx
```diff
@@ -1,45 +1,99 @@
...
+                return;
             }
+
+            setAuthState({
+                state: AUTH_STATES.AUTHENTICATED,
+                user: response.data,
+            });
+        }
+
+        const unsubscribe = auth.onAuthStateChanged(setUser);
```
jaydhulia, line 64: `unsubscribe`? What is this line's purpose?

jaydhulia (approved):
Well, this isn't really solving the race condition issue at its root because I still see the first 403 request to `user_profile` being made before the call to `admin_login` is complete. It "fixes" it by just making a request again later. IMO not the best fix because the root problem is still not fixed, but oh well. 

Here is the order of requests I see, when I hit the login button:
image
That red `user_profile` that 403s shouldn't be in there ideally, because it should have waited fr the `admin_login` POST call to finish first. This works because you are just making a duplicate call again right after, but again that's not ideal since you still have the original race condition.

Approving for now though, as it seems this workaround should work for now.

src/Providers/UserProviders.jsx
ahmedhamodi: Probably not the best name, but essentially it sets up a listener so when auth.onAuthStateChanged happens, the `setUser` function is called (which is defined above on line 30).

src/Providers/UserProviders.jsx
```diff
@@ -1,45 +1,99 @@
...
+                return;
             }
 
-            // validate cookies are set appropriately on this user instance
-            try {
-                const response = await httpGet(baseEndpoint + "user_profile");
-                this.setState({ isAdmin: response.status === 200 });
-            } catch {
-                this.setState({ isAdmin: false });
+            const response = await httpGet(baseEndpoint + "user_profile");
```
ahmedhamodi, line 48: re: comment about race condition still being there

On login this useEffect gets run and this specific httpGet call errors out, then admin_login completes, and then a call is forced using reloadUser (see function on line 71).

One work around is to nest this call in a check that the cookies dcc_auth is set (I can do this with a cookies library) but I didn't want to add another dependency if not absolutely needed.

jaydhulia: Well, checking if the cookie is set doesn't solve it either. Users could have a cookie set, but it's invalid because it is expired. 
So here the problem is, you are parallelizing the call to admin_login and httpGET to user_profile. But in reality, you should be awaiting for the call to admin_login to end before calling user_profile, since there's no point doing this call if admin_login hasn't returned. So if you wanted to fix this, you would make changes to fix this specific ordering of calls.

Don't worry about this, as I said, let's get this merged in, been too long.

PR #97, opened 2021-07-08
Model Viewer Asset Screenshots

vivianliu0:
Loom Demo:
https://www.loom.com/share/1df692086af24e1aa36eae08e3fa6e03

### DIFF
--- a/src/Components/Admin/Assets/assetCard.js
+++ b/src/Components/Admin/Assets/assetCard.js
@@ -79,7 +79,7 @@ export default function AssetCard({ data, handleDeleteClick }) {
             >
                 <img
                     className={classes.cardImage}
-                    src={data.image ? data.image : defaultImage}
+                    src={data.screenshot_url ? process.env.REACT_APP_ADMIN_ASSET_PREFIX + data.screenshot_url : defaultImage}
                     alt="Asset"
                 />
             </Grid>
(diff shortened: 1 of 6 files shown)


src/Components/Admin/Assets/assetCard.js
```diff
@@ -79,7 +79,12 @@ export default function AssetCard({ data, handleDeleteClick }) {
             >
                 <img
                     className={classes.cardImage}
```
ahmedhamodi, line 81: Given the different heights/widths of the default image and of the screenshot, let's add an attribute to cardImage.

If you add: `objectFit: contains` or `objectFit: covers` (please test which one looks better locally). You can find more about object-fit property here: https://www.w3schools.com/css/css3_object-fit.asp

Can you also retroactively add this single prop to the other cards (scenarioCard and sceneCard)? Thanks :)

ahmedhamodi: Wanted to add an image to show you what I mean. The armchair is too tall, so using object-fit should make it look better. 
image

jaydhulia (approved):
This LGTM!

ahmedhamodi:
I assume there's nothing we can do about assets that load side ways right @vivianliu0?

Ie. this sunframe3
image

vivianliu0:
> I assume there's nothing we can do about assets that load side ways right @vivianliu0?
> 
> Ie. this sunframe3
> image

Yeah :/ not sure we can do anything about that

ahmedhamodi (approved):
lgtm, tested locally :)

PR #98, opened 2021-07-10
Puzzle CRUD: Unordered Puzzle

amoliksingh:
Loom video: https://www.loom.com/share/24b293a808b14530bc1d981298d05a21

### DIFF
--- a/src/Components/Admin/ObjectEditor/object.js
+++ b/src/Components/Admin/ObjectEditor/object.js
@@ -2,11 +2,12 @@ import React, { useEffect, useState } from "react";
 import { makeStyles } from "@material-ui/core/styles";
 import { useErrorHandler } from "react-error-boundary";
 import Select from "react-select";
-import TextPaneView from "../ObjectEditor/textpaneview";
 import { Button, IconButton } from "@material-ui/core";
 import AddIcon from "@material-ui/icons/Add";
 import { DeleteForever } from "@material-ui/icons";
+import TextPaneView from "../ObjectEditor/textpaneview";
 import VisualPaneView from "../ObjectEditor/visualpaneview";
+import UnorderedPuzzle from "../ObjectEditor/unorderedpuzzle";
 
 import { getPuzzle, editPuzzle } from "../../../lib/puzzleEndpoints";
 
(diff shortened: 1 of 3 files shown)


src/Components/Admin/ObjectEditor/object.js
```diff
@@ -116,6 +132,15 @@ export default function ObjectEditor({
                 } else if (obj.value === "visual-pane") {
                     animCopy.blackboardData.jsonData.position = [0, 0, 0];
                     animCopy.blackboardData.jsonData.scaleBy = 10;
+                } else if (obj.value === "unordered-puzzle") {
+                    animCopy.blackboardData.componentType = "ordered-puzzle";
```
ahmedhamodi, line 136: This seems wrong?

If condition has: `=== "unordered-puzzle"`
But you set the componentType to `ordered-puzzle`

amoliksingh: In the backend, there isn't actually a "unordered-puzzle" type, it's just "ordered-puzzle" with useTargets set to False

ahmedhamodi:
@amoliksingh - Small issue when I go from higher image count to lower image count.

Steps to reproduce:
- select 3 images for unordered puzzle
- upload 3 images
- go down to 2 images for unordered puzzle
- go back to 3 (or 4/5) images
- the third one will show that a file is uploaded (but not show the preview)

See image below: 
image

What I assume is happening is that the image won't actually show up on the user side, but the file names array isn't getting cleared properly on # of image change.

amoliksingh:
> @amoliksingh - Small issue when I go from higher image count to lower image count.
> 
> Steps to reproduce:
> 
> * select 3 images for unordered puzzle
> * upload 3 images
> * go down to 2 images for unordered puzzle
> * go back to 3 (or 4/5) images
> * the third one will show that a file is uploaded (but not show the preview)
> 
> See image below:
> image
> 
> Expected behaviour:
> When changing to 2 images, 3rd image should be completely wiped from memory. So if I were to save at 2, that image is gone and won't show up. If I change back to 3, 4, or 5 that image won't show up anywhere.
> 
> What I assume is happening is that the image won't actually show up on the user side, but the file names array isn't getting cleared properly on # of image change.

Images are getting cleared when going from 3 --> 2 --> 3+, there was a small bug affecting the uploaded state variable, should be fixed now

ahmedhamodi (approved):
Works great locally. Thanks for getting this done

PR #103, opened 2021-07-21
Puzzle Crud: Jigsaw Puzzle

amoliksingh:
Loom Link: https://www.loom.com/share/9f8088c76e884f9ca5326baea5515ceb


ahmedhamodi (commented):
Haven't looked at code yet, but for the loom video I feel like the preview can be better. Is there a way to set the width to be "auto" or something dynamic so it takes up the least space possible? Perhaps moving away from material-ui's grid is needed. But can you look into that a bit more and see if it's possible? Looks odd atm to me

amoliksingh:
> Haven't looked at code yet, but for the loom video I feel like the preview can be better. Is there a way to set the width to be "auto" or something dynamic so it takes up the least space possible? Perhaps moving away from material-ui's grid is needed. But can you look into that a bit more and see if it's possible? Looks odd atm to me



> Haven't looked at code yet, but for the loom video I feel like the preview can be better. Is there a way to set the width to be "auto" or something dynamic so it takes up the least space possible? Perhaps moving away from material-ui's grid is needed. But can you look into that a bit more and see if it's possible? Looks odd atm to me

Updated Loom Link

ahmedhamodi:
Hm I don't know if this solves the issue. If you upload a very horizontally long image, what happens?

amoliksingh:
> Hm I don't know if this solves the issue. If you upload a very horizontally long image, what happens?

image
At the moment, the issue is fixed, but for small pictures, the spacing is wide.
This shouldn't matter since in a later PR, the image preview is going to change to a single image instead of a grid with 9 images from S3

ahmedhamodi (approved):
Conditional approval. Current issues (same with all puzzle CRUD):
- Image previews are loaded from S3, which means images are being uploaded prior to save
- The way image preview works needs to be reworked to use the image blob uploaded
- Saving to S3 happens on "save" button press -> will need to store multiple images to upload depending on puzzle type

I will create a separate ticket for this, and Amolik will implement after finishing keypad/keyboard puzzle type.

PR #105, opened 2021-07-21
Allow image upload on scenario creation

ahmedhamodi:
*NEW* Loom demo (aug 3): https://www.loom.com/share/1a9dcb4e069b42f98d09400f7d4893bb

Associated backend PR: https://github.com/uwblueprint/cdc-backend/pull/149

Last todo:
- [x] Get preview working with the image directly, no s3 link

### DIFF
--- a/src/Components/Admin/Rooms/roomCard.js
+++ b/src/Components/Admin/Rooms/roomCard.js
@@ -171,7 +171,11 @@ export default function RoomCard({ data, handleEditClick, handleDeleteClick }) {
                                 ? classes.selectedCardImage
                                 : classes.cardImage
                         }
-                        src={data.image ? data.image : defaultImage}
+                        src={
+                            data.screenshot_url
+                                ? data.screenshot_url
+                                : defaultImage
+                        }
                         alt="Escape Room"
                     />
                 </Grid>
--- a/src/Components/Admin/Rooms/roomCard.js
+++ b/src/Components/Admin/Rooms/roomCard.js
@@ -172,8 +172,9 @@ export default function RoomCard({ data, handleEditClick, handleDeleteClick }) {
                                 : classes.cardImage
                         }
                         src={
-                            data.screenshot_url
-                                ? data.screenshot_url
+                            data.display_image_url
+                                ? process.env.REACT_APP_ADMIN_ASSET_PREFIX +
+                                  data.display_image_url
                                 : defaultImage
                         }
                         alt="Escape Room"
--- a/src/Components/Admin/Rooms/roomCard.js
+++ b/src/Components/Admin/Rooms/roomCard.js
@@ -121,6 +121,7 @@ export default function RoomCard({ data, handleEditClick, handleDeleteClick }) {
     const history = useHistory();
     const [anchorEl, setAnchorEl] = React.useState(null);
     const [isHighlighted, setIsHighlighted] = React.useState(false);
+    const imageHash = Date.now();
 
     const handleMenuClick = (event) => {
         setAnchorEl(event.currentTarget);
@@ -166,6 +167,7 @@ export default function RoomCard({ data, handleEditClick, handleDeleteClick }) {
                     justify="center"
                 >
                     <img
+                        key={data.id}
                         className={
                             isHighlighted
                                 ? classes.selectedCardImage
@@ -174,7 +176,9 @@ export default function RoomCard({ data, handleEditClick, handleDeleteClick }) {
                         src={
                             data.display_image_url
                                 ? process.env.REACT_APP_ADMIN_ASSET_PREFIX +
-                                  data.display_image_url
+                                  data.display_image_url +
+                                  "?" +
+                                  imageHash
                                 : defaultImage
                         }
                         alt="Escape Room"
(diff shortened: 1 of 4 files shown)


src/Components/Admin/Rooms/roomModal.js
```diff
@@ -60,6 +73,31 @@ export default function RoomModal({
...
+            const split = name.split(lastDot);
+            const fileName = split[0];
+            const fileType = split[1];
+
+            setFileName(fileName);
+            handleUploadAssetSubmit(fileName, fileType, file);
+        }
+    };
+
+    const handleUploadAssetSubmit = async (name, file_type, asset) => {
```
jaydhulia, line 91: Bad copypasta here too, the name should be fixed to reflect this is not an asset upload

src/Components/Admin/Rooms/roomModal.js
```diff
@@ -60,6 +73,31 @@ export default function RoomModal({
...
+        const response = await createPresignedLinkAndUploadS3(
+            { file_type: file_type, type: "asset", file_content: asset },
```
jaydhulia, line 93: Again this is not correct: the type shouldn't be an `asset`. It will store the image in the wrong prefix in S3. This should be type `image`. See Amolik's implementation for this: https://github.com/uwblueprint/cdc-frontend/blob/a8f1e94b58e04c5cfb0ffafe15690da554431b44/src/Components/Admin/ObjectEditor/visualpaneview.js#L44

src/Components/Admin/Rooms/roomCard.js
```diff
@@ -81,7 +81,9 @@ export default function RoomCard({ data, handleEditClick, handleDeleteClick }) {
             >
                 <img
                     className={classes.cardImage}
-                    src={data.image ? data.image : defaultImage}
+                    src={
+                        data.screenshot_url ? data.screenshot_url : defaultImage
```
jaydhulia, line 85: You are saving the full screenshot URL in the backend, instead of just the key (see backend PR comment: https://github.com/uwblueprint/cdc-backend/pull/149#discussion_r673629733). So once you fix that in the backend PR, I believe you will have to change the screenshot_url here to prepend the s3 prefix.

jaydhulia (changes requested):

src/Components/Admin/Rooms/roomModal.js
ahmedhamodi: Sloppy. My bad!

aaronabraham311: Correct me if I'm wrong, but I thought we generalized the upload functions so we don't need to keep creating these functions every time we need to upload to S3

aaronabraham311 (commented):
Codewise looks fine. The image is cut-off when I see it in the Loom video for the scene cards. Can we change the scene cards to fit the image rather than scale and cut?

src/Components/Admin/Rooms/roomModal.js
ahmedhamodi: It is generalized! The `createPresignedLinkAndUploadS3` function does all the behind the scenes logic. The parent function just calls it and sets appropriate values on the response from this function call.

ahmedhamodi:
> The image is cut-off when I see it in the Loom video for the scene cards.

Assuming you mean scenario cards? And sure. A couple of issues around how the image will fit though, if we fit the whole width the height is going to be weird. Three options for height:
- stretch image (no go for me)
- center image horizontally and have white space on top and bottom (also not ideal imo)
- maintain aspect ratio + no centering, but then each card has different height (not ideal imo)

Unless there's a better option, I guess I can proceed with centering image horizontally, but I'm not a huge fan of any of these.

ahmedhamodi:
Option 1:
image

Option 2: 
image

Option 3:
image

Thoughts @aaronabraham311?

aaronabraham311:
Oh wow they all have drawbacks. Perhaps option 3 is best but that changes card sizes.

ahmedhamodi:
> Oh wow they all have drawbacks. Perhaps option 3 is best but that changes card sizes.

Yeah exactly 😦 , not sure what is best here.

aaronabraham311 (approved):
Let's keep it as is then. Overall, I have no other issues

src/Components/Admin/Rooms/roomModal.js
```diff
@@ -48,6 +60,8 @@ export default function RoomModal({
         setIsPublished(room ? room.is_published : false);
         setIsPreviewable(room ? room.is_previewable : false);
         setRoomSolveTime(room ? room.expected_solve_time : "");
+        setPreviewSrc(room ? room.display_image_url : "");
+        setFileName(room ? room.display_image_url : "");
```
jaydhulia, line 64: Shouldn't this be ""?

ahmedhamodi:
Newest loom video: https://www.loom.com/share/1a9dcb4e069b42f98d09400f7d4893bb

Shows s3 maintenance as expected.

@amoliksingh you will want to do a similar thing for puzzle CRUD when you get around to that.

aaronabraham311 (approved):
:ship:

PR #106, opened 2021-07-24
Puzzle CRUD: Keypad Puzzles

amoliksingh:
Loom video: https://www.loom.com/share/8ffbfd888e0f4e229c125dd9cd5b23ef

### DIFF
--- a/src/Components/Admin/ObjectEditor/keypadpuzzle.js
+++ b/src/Components/Admin/ObjectEditor/keypadpuzzle.js
@@ -1,28 +1,73 @@
 import React, { useState } from "react";
 import AddIcon from "@material-ui/icons/Add";
-import { IconButton } from "@material-ui/core";
+import { Button, IconButton } from "@material-ui/core";
 import { DeleteForever } from "@material-ui/icons";
 import TextField from "@material-ui/core/TextField";
+import { makeStyles } from "@material-ui/core/styles";
+
+const useStyles = makeStyles(() => ({
+    input: {
+        "&:invalid": {
+            border: "red solid 2px",
+        },
+    },
+}));
 
 export default function KeypadPuzzle(props) {
+    const classes = useStyles();
+    const [tempPass, setTempPass] = useState(props.pass);
     const [pass, setPass] = useState(props.pass);
+    const [showNumpadPrompt, setNumpadPrompt] = useState(false);
+    const [showKeyboardPrompt, setKeyboardPrompt] = useState(false);
+    const [showSavePass, setSavePass] = useState(false);
 
-    const addPass = () => {
-        const newPass = prompt("Enter the text for the puzzle: ");
-
-        if (newPass) {
-            if (newPass !== /^[0-9.,]+$/) {
-                alert("Error: Password may only contain numbers");
-            } else {
-                props.savePass(newPass);
-                setPass(newPass);
-            }
+    const showPrompt = () => {
+        if (props.isNumpad) {
+            setNumpadPrompt(true);
+        } else {
+            setKeyboardPrompt(true);
         }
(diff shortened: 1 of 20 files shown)


src/Components/Admin/ObjectEditor/keypadpuzzle.js
```diff
@@ -0,0 +1,121 @@
...
+
+    const numpadOnChange = (e) => {
+        const re = /^[0-9\b]{1,7}$/;
+        if (e.target.value === "" || re.test(e.target.value)) {
+            setTempPass(e.target.value);
+            setSavePass(true);
+        } else {
+            alert(
+                "Error: Password must be at most 7 characters and only consist of numbers"
+            );
```
ahmedhamodi, line 57: Todo in future: remove alerts and use dynamic html elements to show error. ie. have an error boolean, have some red text hidden when boolean is false, and show when it is true. instead of alert here, we toggle the boolean.

src/Components/Admin/ObjectEditor/keypadpuzzle.js
```diff
@@ -0,0 +1,121 @@
...
+
+    const keyboardOnChange = (e) => {
+        const re = /^[0-9a-zA-Z\b]{1,7}$/;
+        if (e.target.value === "" || re.test(e.target.value)) {
+            setTempPass(e.target.value);
+            setSavePass(true);
+        } else {
+            alert(
+                "Error: Password must be at most 7 characters and must be alphanumeric"
+            );
```
ahmedhamodi, line 70: same here

src/Components/Admin/ObjectEditor/keypadpuzzle.js
```diff
@@ -0,0 +1,121 @@
...
+        } else {
+            alert(
+                "Error: Password must be at most 7 characters and only consist of numbers"
+            );
+            setSavePass(false);
+        }
+    };
+
+    const keyboardOnChange = (e) => {
+        const re = /^[0-9a-zA-Z\b]{1,7}$/;
```
ahmedhamodi, line 63: I think keyboard password can be longer than 7 characters (I see no spacial reason on the keyboard interface to indicate otherwise). However when testing currently, I was only allowed to enter 5 (likely because expected password is "abcde" which is 5 characters). We should probably allow users to input more than the expected password length, if that is possible to do. @jaydhulia or @drbalar can you verify if that is possible or not?

Image:
image

ahmedhamodi (commented):
Code looks good besides the few comments and future todos I outlined. Works locally to (I'm not a huge fan of the user experience on entering an invalid password (the alert + very different css on the input field) but we can fix that later.

Will approve after we figure out the regex issue on keypad/keyboard.

src/Components/Admin/ObjectEditor/keypadpuzzle.js
drbalar: Yea it is possible to remove the password length restriction

amoliksingh: Set max length to 16 as per recommendation

src/Components/Admin/ObjectEditor/keypadpuzzle.js
```diff
@@ -0,0 +1,122 @@
...
+        if (props.isNumpad) {
+            setNumpadPrompt(true);
+        } else {
+            setKeyboardPrompt(true);
+        }
+        setSavePass(true);
+    };
+
+    const deletePass = () => {
+        alert(pass);
```
ahmedhamodi, line 34: Intentional alert here?

ahmedhamodi (approved):

PR #107, opened 2021-08-04
Rework delete popup + update environment on scene deletion

ahmedhamodi:
Loom video: https://www.loom.com/share/3080be2d8fbd41028bc8fbc6901cc717

### DIFF
--- a/src/Components/Admin/EnvironmentEditor/environment.js
+++ b/src/Components/Admin/EnvironmentEditor/environment.js
@@ -289,6 +289,9 @@ export default function EnvironmentEditor({
         setScenes(modifiedScenes);
         setDeleteSceneId(null);
         setDeleteModalOpen(false);
+
+        const data = await getScenario(environmentId, handleError);
+        setEnvironment(data);
     };
 
     return (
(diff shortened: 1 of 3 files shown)


src/Components/Admin/EnvironmentEditor/environment.js
```diff
@@ -451,7 +454,10 @@ export default function EnvironmentEditor({
             />
             <DeleteModal
                 open={deleteModalOpen}
-                confirmMessage="Are you sure you want to delete this scene?"
+                title="Delete Scene and Transition"
+                confirmMessage={
+                    "Are you sure you want to delete this scene?\nDeleting the scene will also delete the transition."
```
aaronabraham311, line 459:
```suggestion
                    "Are you sure you want to delete this scene?\nDeleting the scene will also delete the following transition."
```

src/Components/Admin/common/deleteModal.js
```diff
@@ -16,16 +16,28 @@ const useStyles = makeStyles(() => ({
     },
 }));
 
+function NewlineText(string) {
+    return string.split("\n").map((str, index) => (
+        <p style={{ fontSize: 14, lineHeight: "16px" }} key={index}>
+            {str}
+        </p>
+    ));
+}
```
aaronabraham311, line 25: What's this for?

ahmedhamodi: It's so that we can include `\n` characters in the single prop field for the popup and it will display in different paragraphs. It is being used in the delete scene popup (the extra information about deleting transitions is in a new paragraph). It looks better imo. Generalized the logic so other popups can use this too if needed.

aaronabraham311 (approved):

PR #108, opened 2021-08-16
Puzzle CRUD: Fix image logic

amoliksingh:
Fixed it to only upload to S3 after clicking save and to reuse the same s3 keys
Loom video: https://www.loom.com/share/eeaab7398a1a466493ce9ac8f25241ca
The jigsaw puzzle bug is fixed by Jay's PR


ahmedhamodi (approved):
Looks great! Thanks for getting this done

PR #112, opened 2021-10-16
Puzzle CRUD: Rework unordered puzzle logic

amoliksingh:
Loom video: https://www.loom.com/share/3a842811226745e2885fde9cafc3ce06

Basically, this PR reworks the unordered puzzle to use add and delete, and also ensures that when deleting images, they are in fact also deleted from S3

### DIFF
--- a/src/Components/Admin/ObjectEditor/unorderedpuzzle.js
+++ b/src/Components/Admin/ObjectEditor/unorderedpuzzle.js
@@ -167,14 +167,13 @@ export default function UnorderedPuzzle(props) {
         <div>
             {isUnordered ? (
                 <div>
-                    Add Images
+                    Add Images ({images.length}/5)
                     <IconButton
                         className={props.classes.addButton}
                         aria-label="add"
+                        disabled={images.length === 5}
                         onClick={() => {
-                            if (images.length !== 5) {
-                                addImage();
-                            }
+                            addImage();
                         }}
                     >
                         <AddIcon />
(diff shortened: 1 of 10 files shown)


ahmedhamodi:
Is this good for me to review @amoliksingh?

amoliksingh:
> Is this good for me to review @amoliksingh?

Yes

src/Components/Admin/ObjectEditor/unorderedpuzzle.js
```diff
@@ -158,18 +149,35 @@ export default function UnorderedPuzzle(props) {
...
+                    <IconButton
+                        className={props.classes.addButton}
+                        aria-label="add"
+                        onClick={() => {
+                            addImage();
+                        }}
+                    >
+                        <AddIcon />
+                    </IconButton>
+                </div>
```
ahmedhamodi, line 180: So you know how there is a disabled prop on the delete button when # images == 2, should we have a disabled prop on add button when # images == 5?

ahmedhamodi (commented):
Code looks good, only 1 question. Will test out locally now

src/Components/Admin/ObjectEditor/unorderedpuzzle.js
amoliksingh: Good idea, I think we probably should

ahmedhamodi: Can you add that and I'll re-review

amoliksingh: Added

ahmedhamodi (approved):
LGTM

PR #122, opened 2021-12-08
Style scene card dropdown menu

ahmedhamodi:
Loom video: https://www.loom.com/share/e662438d18fe41a9a45b0004e8a61466

Please watch the video as I brought up a concern I had with the actual values in this dropdown, for now 4/5 are disabled because they do not do anything, so was wondering if there even is a use case for them moving forward.


drbalar (approved):
Loosk good

PR #127, opened 2021-12-18
Add is_previewable toggle + disabled this modification in Edit modal

ahmedhamodi:
Is published and is previewable is view only on edit panel (since we have designated panel to modify it):
image

The designated panel:
image


drbalar (approved):
LG!

PR #128, opened 2021-12-18
Added ability to upload images to transitions

drbalar:
Loom: https://www.loom.com/share/930b10ada7874d7c9965caebdf29dba4

### DIFF
--- a/src/Components/Admin/EnvironmentEditor/environment.js
+++ b/src/Components/Admin/EnvironmentEditor/environment.js
@@ -25,6 +25,7 @@ import {
 
 import "../../../styles/index.css";
 import { Colours } from "../../../styles/Constants.ts";
+import { createPresignedLinkAndUploadS3 } from "../../../lib/s3Utility";
 
 const useStyles = makeStyles((theme) => ({
     page: {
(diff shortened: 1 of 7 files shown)


ahmedhamodi:
If you can leave a comment in your PR when you have validated the following edge cases, I'll approve:
1. upon save and reupload of an image, the previous image is no longer in AWS (either by overriding or deleting and reuploading, depending on file type mismatch)
2. upon deletion of a transition, the image is also deleted if it exists

Thanks!

src/Components/Admin/EnvironmentEditor/environment.js
```diff
@@ -282,8 +286,39 @@ export default function EnvironmentEditor({
...
+                            const imageList = [oldS3Key];
+                            await deleteTransitionImages(
+                                {
+                                    scenarioId: environmentId,
+                                    imagesList: imageList,
+                                },
+                                handleError
+                            );
+                        }
+                    }
```
ahmedhamodi, line 336: I may need this logic explained to me in more detail.

My understanding:
If image exists, find the old S3 key and the old file type (so this can be the S3 key and .jpg let's say).
Issue: what if I delete the image/transition? the transitions imageSrc will be empty, so this conditional isn't run, and no file deletion is happening on S3 side right? Unless I am missing something.

Next, if file type is the same, set s3Key to the old s3Key. What happens now? Are we uploading to AWS the new image, and reusing the key value?

If file type is not the same, we delete image. What happens here? Do we still upload a new image if I have selected a new image?

drbalar: So there was some caching stuff going on with using the same link. What we now do to tackle this is delete the image regardless and upload the new image to a new key.

We maintain a list of images to delete and then at the end when we click submit, the images on that list are executed 😵😵. Lmk if you have questions about the new approach.

ahmedhamodi (approved):
Makes sense now, thanks!

PR #132, opened 2022-01-11
New images for default room and default asset thumbnails

ahmedhamodi:
Loom demo: https://www.loom.com/share/a450b555c7d446ebaed1c05540f2e54f


amoliksingh (approved):

jaydhulia:
Looks good to me! For the scene pic, we are getting rid of the scene tab, but eventually, we will still have a scene image (inside the room view where the transitions are). So eventually we will need a scene image, but that can be done in a later PR when we implement pics for the scene inside the room view.

ahmedhamodi:
Sounds good Jay.

PR #133, opened 2022-01-11
Bug fix: only PUT request when change on share & publish modal

ahmedhamodi:
Loom video: https://www.loom.com/share/9c9fa0d6014f4b04b639b885f829f2ae


src/Components/Admin/Rooms/roomModal.js
```diff
@@ -261,7 +261,11 @@ export default function RoomModal({
     };
 
     const handleModalCloseClick = () => {
-        if (isShareAndPublish) {
+        if (
```
jaydhulia, line 264: Should we have a toast message or something to show that the changes were saved? 
I guess we currently don't do that right now, so not a biggie.

jaydhulia (approved):

jaydhulia:
Looks good to me. Minor UX comment.

src/Components/Admin/Rooms/roomModal.js
ahmedhamodi: Definitely a nice to have throughout the app in the future. If we have time, I'll def do this!

PR #140, opened 2022-01-20
Update styling for transition modal

drbalar:
https://www.loom.com/share/e04718eeceae4e10900751c8c3ec035a

changes made:
- choose file button is more colorful
- preview images are limited to max height of 200px and max width of 500 px
- Add link button replaced with a text field (to avoid the chrome prompt). Also added a `Test URL` button for testing the link

I put the css in the index.css file. I wasn't quite sure where's the right spot for it so if there is a better spot, lmk and i can move it


ahmedhamodi (approved):
Looks great! Side effect of this change: the input buttons for escape room creation + asset upload also changed due to the changes in index.css. It looks good though, just wanted to bring this up so you are aware there are more global impacts to this change
image

drbalar:
> Looks great! Side effect of this change: the input buttons for escape room creation + asset upload also changed due to the changes in index.css. It looks good though, just wanted to bring this up so you are aware there are more global impacts to this change
> image

Ohh good point, I forgot about that part.. thanks for testing it out!

ahmedhamodi:
No problem :)

PR #141, opened 2022-01-25
Puzzle CRUD UI Changes 2

amoliksingh:
Figma Link: https://www.figma.com/file/XcNmjKSIsf6Q1CxpGDVtL2/High-Fidelity?node-id=632%3A12172
image
image
image
image


ahmedhamodi (approved):
LGTM

ahmedhamodi:
I do think the design itself (ie, the button "Change File" location, padding around it, etc) needs to be improved but I assume that will be done in a future PR. The purpose of this PR is to conditionally render the button when uploaded, which it does.

PR #142, opened 2022-01-29
[Bug Fix] Keep user logged in on new page if auth token exists

ahmedhamodi:
Loom video: https://www.loom.com/share/ecb2498212b54e97a34bfc336fb89910

### DIFF
--- a/src/Routes.js
+++ b/src/Routes.js
@@ -1,4 +1,4 @@
-import React, { useContext, useEffect } from "react";
+import React, { useContext, useEffect, useState } from "react";
 import { Route, Switch } from "react-router-dom";
 import Login from "./Components/Login/login";
 import Signup from "./Components/Login/signup";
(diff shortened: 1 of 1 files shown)


amoliksingh (approved):
LGTM, I logged in at http://localhost:3000/admin/environment/1 and it redirected me to http://localhost:3000/admin, is this expected? Also not sure about the failed request either mentioned in the loom video

ahmedhamodi:
> I logged in at http://localhost:3000/admin/environment/1 and it redirected me to http://localhost:3000/admin, is this expected?
I don't believe we added any logic to save the route that was attempted at login, but this is something I can explore as I know it was done in some other repos I work on. Will do outside of this PR though.

> Also not sure about the failed request either mentioned in the loom video
Yeah that is fair, I'm not sure if there is a "better" way to do it that isn't hacky and looking for the existence of the exact token name before making the call, which I think isn't proper?

Will merge this as is, but happy to change in the future if we decide it is not appropriate to be doing it the way I am.

PR #145, opened 2022-02-08
Add share and publish modal in environment editor

ahmedhamodi:
Loom demo: https://www.loom.com/share/91eff1b010d64c5482d260633b10d803

### DIFF
--- a/src/Components/Admin/Rooms/roomModal.js
+++ b/src/Components/Admin/Rooms/roomModal.js
@@ -84,6 +84,7 @@ export default function RoomModal({
     room,
     isEdit,
     isShareAndPublish,
+    isEnvBar,
 }) {
     const classes = useStyles();
     const handleError = useErrorHandler();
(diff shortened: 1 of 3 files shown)


drbalar:
there an issue im running into when testing this out.. not sure exactly but maybe something to do with the states? 
Here's a loom attempting to explain it but lmk if you have more questions
https://www.loom.com/share/2eff7269aae3479bb5489de6077c6c55

ahmedhamodi:
Yeah that appears to be a valid bug. Will investigate tomorrow! Thanks for catching.

src/Components/Admin/Rooms/roomModal.js
```diff
@@ -341,7 +344,9 @@ export default function RoomModal({
             });
         }
 
-        resetFields();
+        if (!isEnvBar) {
+            resetFields();
+        }
```
ahmedhamodi, line 349: Fixed the bug. The issue was that on home page when you open share and publish, the room is being set which redefines the fields and resets the fields on each modal close. However once in the environment editor, the room isn't being set on each open so we cannot reset. This change makes sure that doesn't happen.

Ready for retesting and approval now @drbalar

PR #146, opened 2022-02-08
Enable play button in environment editor

ahmedhamodi:
The play button in the top right of environment editor:
image

When clicked will open a new tab with the student experience open (ie. http://localhost:8888/youth-in-crisis for local environment).


drbalar (approved):
looks good

PR #150, opened 2022-02-10
Edit Hints Modal

ahmedhamodi:
Loom demo:
https://www.loom.com/share/93603e0ddb214957b7ac3f6ee44b5bf7

Blocker: Validation on hints is very restrictive, doesn't work on any of the hints that are already made.


jaydhulia:
image
Hitting "Edit scene hits" and then hitting cancel on modal both are causing calls to BE to the assets endpoint... seems like a bug?

ahmedhamodi:
Fixed @jaydhulia. I was reusing a variable since I was trying to be smart with variable usage, but then any changes to that variable caused a reload in the SceneModal which calls `assets` to get background assets to use when making a new scene.

drbalar:
@ahmedhamodi Came across this one issue https://www.loom.com/share/c2b9f287e43d436693e5306cd594148e
Besides this, it all looking good

ahmedhamodi:
Will fix tomorrow!

PR #151, opened 2022-02-16
Ability to edit transitions

ahmedhamodi:
Loom demo: https://www.loom.com/share/5b26337464e14ba9a7b8476401977a52

### DIFF
--- a/src/Components/Admin/EnvironmentEditor/transitionModal.js
+++ b/src/Components/Admin/EnvironmentEditor/transitionModal.js
@@ -311,7 +311,7 @@ export default function TransitionModal({
                                             padding: 10,
                                         },
                                     }}
-                                    placeholder="Hint text"
+                                    placeholder="Enter transition text"
                                     multiline
                                 />
                             </div>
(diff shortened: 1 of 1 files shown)


src/Components/Admin/EnvironmentEditor/transitionModal.js
```diff
@@ -266,21 +286,35 @@ export default function TransitionModal({
...
+                                    onChange={(e) => handleTextChange(e, index)}
+                                    className={classes.textField}
+                                    required
+                                    variant="outlined"
+                                    inputProps={{
+                                        style: {
+                                            padding: 10,
+                                        },
+                                    }}
+                                    placeholder="Hint text"
```
drbalar, line 314: I think this should be transistions text or something?

ahmedhamodi: Yup you're right! Fixed.

drbalar (approved):
Lookin good 🚀

PR #155, opened 2022-03-02
Fix asset uploading

ahmedhamodi:
Loom demo: https://www.loom.com/share/152bff0092f64aee8a64a92a10e708b5

### DIFF
--- a/src/Components/Admin/Assets/fileTypes.ts
+++ b/src/Components/Admin/Assets/fileTypes.ts
@@ -1,5 +1,5 @@
 export enum FileTypes {
-    GLTF = "gltf",
-    GLB = "glb",
+    GLTF = ".gltf",
+    GLB = ".glb",
     NONE = "none",
 }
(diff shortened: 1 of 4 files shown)


src/Components/Admin/Assets/fileTypes.ts
```diff
@@ -1,5 +1,5 @@
 export enum FileTypes {
-    GLTF = "gltf",
-    GLB = "glb",
+    GLTF = ".gltf",
+    GLB = ".glb",
```
ahmedhamodi, line 3: I made this change so I can use these directly as file type restrictions on line 254 in uploadAssetModal.js

amoliksingh (commented):
https://www.loom.com/share/730aee178d844fe79d05adda7a1ca437
Weird error happening for gltf file

src/Components/Admin/Assets/fileTypes.ts
jaydhulia: BE already adds a dot to the extension, so FE just needs to send gltf or glb, rather than .gltf or .glb. That's the cause of the error that @amoliksingh posted in his loom video

ahmedhamodi:
> Weird error happening for gltf file

Fixed!

drbalar (approved):
LG 💯

PR #157, opened 2022-03-02
Add external link to conclusion editor

ahmedhamodi:


src/Components/Admin/EnvironmentEditor/conclusionModal.js
```diff
@@ -126,13 +135,38 @@ export default function ConclusionModal({
...
+                        onChange={handleExternalLinkChange}
+                        required
+                        variant="outlined"
+                        multiline
+                        inputProps={{
+                            style: {
+                                padding: "0px 10px",
+                            },
+                        }}
+                        placeholder="The external link students will can go to after the room is done."
```
jaydhulia, line 159: Small typo. Would suggest changing to:
"The feedback form link students can use to submit feedback" 
Because the external_link attribute is only used for feedback forms.

src/Components/Admin/EnvironmentEditor/conclusionModal.js
```diff
@@ -126,13 +135,38 @@ export default function ConclusionModal({
...
                     />
                 </div>
+                <div style={{ paddingBottom: 15 }}>
+                    <Typography
+                        style={{
+                            fontSize: 20,
+                            lineHeight: "27px",
+                        }}
+                    >
+                        External Link
```
jaydhulia, line 145: Similar to below, would suggest changing to "Feedback Link"

jaydhulia (approved):

PR #158, opened 2022-03-02
Email on report button click

ahmedhamodi:
image

To field is also populated.


src/Components/Admin/common/errorModal.js
```diff
@@ -42,13 +42,21 @@ export default function ErrorModal({ error, resetErrorBoundary }) {
...
             <DialogContent>{message}</DialogContent>
             <DialogActions className={classes.buttonContainer}>
-                <Button
-                    onClick={() => {
-                        // TODO: send email to us
-                    }}
+                <a
+                    href={
+                        "mailto:" +
+                        process.env.REACT_APP_EMAIL +
```
ahmedhamodi, line 48: Note to all of us, when we deploy this we need to add this key to the netlify deployment

drbalar (approved):
LGTM 🪐

PR #162, opened 2022-03-07
Proper save logic on asset model viewer

ahmedhamodi:


ahmedhamodi (commented):
Self reviewing. Merging so we can have it deployed for our testing in today's FYDP work session.

PR #169, opened 2022-03-22
Hide Stats Page

ahmedhamodi:


src/Components/Admin/admin.js
```diff
@@ -582,13 +582,13 @@ export default function Admin() {
                             handleDeleteAssetClick={handleDeleteAssetClick}
                         />
                     </TabPanel>
-                    <TabPanel
+                    {/* <TabPanel
```
ahmedhamodi, line 585: Just commenting it out in case we want to come back to it. Better than removing imo!

drbalar (approved):

PR #171, opened 2022-03-28
Fix password bug and add optional image to texts

amoliksingh:
Fix password bug and add optional image to texts


ahmedhamodi (approved):
LGTM

